PR #2, opened 2017-07-10
Adds a pull request template

JPrevost:
Why:

* :rainbow:

Also adds travis config


thatandromeda:
I infer from the fact that the build passed that your config must work :) 👍

PR #3, opened 2017-07-10
Users

JPrevost:
## Status
**READY**

#### What does this PR do?
This adds an initial User model and configures Devise / Omniauth to provide authentication via
MIT OpenID Pilot (this is a temporary configuration as this project is not authorized to proceed to
production like this, but we need to develop _something_ so this is what we've got for now).

- bundle and migrate database (`bundle install`, `rails db:migrate`)
- set the two `MIT_OAUTH2_API_KEY` and `MIT_OAUTH2_API_SECRET` to valid values in `.env`
- `heroku local`
- localhost:5000/users/auth/mit_oauth2
- login and admire the nice white screen with no context
- `heroku local:run rails console`
- `Users.all` should show your newly created local account

- hey, I should open some tickets (sorry... will do for the next bits of work)

YES

YES


thatandromeda:
If this project has API keys, can they be added to LastPass?

JPrevost:
Yes, there are currently no actual API keys specific to the project. I'll make some and document them. I tested functionality with a non-project specific set of keys (not helpful but the truth).

app/models/user.rb
```diff
@@ -2,15 +2,27 @@
...
-#  admin      :boolean          default(FALSE)
-#  created_at :datetime         not null
-#  updated_at :datetime         not null
+#  id                 :integer          not null, primary key
+#  uid                :string           not null
+#  email              :string           not null
+#  admin              :boolean          default(FALSE)
+#  created_at         :datetime         not null
+#  updated_at         :datetime         not null
+#  sign_in_count      :integer          default(0), not null
```
thatandromeda: Why are we adding all this user-tracking data? Do we need it?

app/views/static/index.html.erb
```diff
@@ -0,0 +1 @@
+Hi!
```
thatandromeda: 👋

config/initializers/devise.rb
```diff
@@ -0,0 +1,278 @@
+# Use this hook to configure devise mailer, warden hooks and so forth.
+# Many of these configuration options can be set straight in your model.
+Devise.setup do |config|
+  # The secret key used by Devise. Devise uses this key to generate
+  # random tokens. Changing this key will render invalid all existing
+  # confirmation, reset password and unlock tokens in the database.
+  # Devise will use the `secret_key_base` as its `secret_key`
+  # by default. You can change it below and use your own secret key.
+  # config.secret_key = '4e69f36dde0bcbc616e46ddb3938978a7f65cc7ef1c4fc47bfed313e7215d12f2288388155bb6a204120ac1cb627ced842f8917bfc7bc73d13dea2773be8ba31'
```
thatandromeda: Does this really want to be in the repo?

app/models/user.rb
JPrevost: It comes along free with Devise. I thought it might be helpful in debugging user problems. If this app had any privacy sensitivity associated I might lean towards not tracking, but since it's "upload your thesis" I think it's legit to have this info for debug purposes. I'm happy to discuss concerns with it though. It is indeed _not_ in the spec ;)

config/initializers/devise.rb
thatandromeda: Do we want our own secret key (which presumably would not live in the repo)?

config/initializers/devise.rb
```diff
@@ -0,0 +1,278 @@
...
+
+  # ==> Mailer Configuration
+  # Configure the e-mail address which will be shown in Devise::Mailer,
+  # note that it will be overwritten if you use your own mailer class
+  # with default "from" parameter.
+  config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
```
thatandromeda: Also do we care about changing this?

config/initializers/devise.rb
JPrevost: That's a commented out fake key. It's all good. Thanks for paying attention though as that would  be super bad to have that for real in the repo.

app/models/user.rb
```diff
@@ -0,0 +1,16 @@
+# == Schema Information
+#
+# Table name: users
+#
+#  id         :integer          not null, primary key
+#  uid        :string           not null
```
thatandromeda: Should this actually be named `kerb` and not `uid` for clarity (and also for ease of not-making-typos-between-`id`-and-`uid`??

config/initializers/devise.rb
JPrevost: We aren't using Devise's mailer as all user's will be created on-the-fly with no email confirmation via the Omniauth integrations.

config/initializers/devise.rb
```diff
@@ -0,0 +1,278 @@
...
+
+  # Configure which authentication keys should be case-insensitive.
+  # These keys will be downcased upon creating or modifying a user and when used
+  # to authenticate or find a user. Default is :email.
+  config.case_insensitive_keys = [:email, :uid]
+
+  # Configure which authentication keys should have whitespace stripped.
+  # These keys will have whitespace before and after removed upon creating or
+  # modifying a user and when used to authenticate or find a user. Default is :email.
+  config.strip_whitespace_keys = [:email, :uid]
```
thatandromeda: (If we change `:uid` to `:kerb` we need to change this too)

app/models/user.rb
JPrevost: I'm not opposed to this suggested change.

config/initializers/devise.rb
```diff
@@ -0,0 +1,278 @@
...
+  # their account can't be confirmed with the token any more.
+  # Default is nil, meaning there is no restriction on how long a user can take
+  # before confirming their account.
+  # config.confirm_within = 3.days
+
+  # If true, requires any email changes to be confirmed (exactly the same way as
+  # initial account confirmation) to be applied. Requires additional unconfirmed_email
+  # db field (see migrations). Until confirmed, new email is stored in
+  # unconfirmed_email column, and copied to email column on successful confirmation.
+  config.reconfirmable = true
```
thatandromeda: If we're requiring reconfirmation of email changes but not configuring mail, will we end up sad?

JPrevost: Sounds like a problem for if we ever allow email changes ;) But good question.

config/initializers/devise.rb
```diff
@@ -0,0 +1,278 @@
...
+  #
+  #     mount MyEngine, at: '/my_engine'
+  #
+  # The router that invoked `devise_for`, in the example above, would be:
+  # config.router_name = :my_engine
+  #
+  # When using OmniAuth, Devise cannot automatically set OmniAuth path,
+  # so you need to do it manually. For the users scope, it would be:
+  # config.omniauth_path_prefix = '/my_engine/users/auth'
+  config.omniauth :mit_oauth2, Rails.application.secrets.oauth_key, Rails.application.secrets.oauth_secret, scope: 'openid,email,profile'
```
thatandromeda: TIL about Rails.application.secrets and it is very cool.

config/initializers/devise.rb
JPrevost: I've waffled in the past on including the stock config for stuff like this and only changing the necessary bits. This exchange has made me sway back towards deleting the entire default config and only putting in the important bits :)

config/initializers/devise.rb
JPrevost: For some reason I often access the ENV directly. I think I'll try to load stuff in through `Rails.application.secrets` on this app and see if we feel more 🌈

thatandromeda:
I would like you to either convince me that the user tracking is merited or delete it, at which point I'll switch this to mergeable - the other questions are left up to your aesthetic judgment.

JPrevost:
I shall make a passing attempt at convincing you of the merits of the login log.

JPrevost:
@thatandromeda 

I'm going to change `User.uid` to `User.kerb_id` (although now I want to change it to `User.kerbal_id`).

I'll make that change and remove the last login bits as I don't feel strongly enough about them to make an argument... other than "one day it'll be helpful to know when someone last logged in when debugging something weird ;)"

app/models/user.rb
JPrevost: noting here for future reference: I was wrong and `uid` is _not_ the kerbid. As such, we are leaving this as `User.uid`.

JPrevost:
@thatandromeda I've removed the tracking bits and made a note on what `User.uid` is in a hopefully semi-useful place. Let me know if this is looking mergeable. If so I'll squash it up and merge it. Thanks!

thatandromeda:
Squash away!

Issue #14, opened 2017-07-19
drop in admin gem


JPrevost:
configure admin gem (not sure which one yet) to allow staff users to use a web ui to manipulate approved data

JPrevost:
closed via #15

Issue #13, opened 2017-07-19
create roles system


JPrevost:
`cancancan` is generally a good fit.

User
- current_user can create a new thesis
- current_user can view and edit their own theses
- can a current_user delete their own theses? (check with stakeholders)
- current_user can create / edit their own advisors

Another User
- cannot create, view, edit or delete another user thesis
-  cannot create, view, edit or delete another user advisors

Staff
- staff can view, edit and delete all thesis
- staff can create / edit departments (not delete!)
- staff can create / edit rights (not delete!)
- staff can create / edit degrees (not delete!)
- staff can create / edit user advisors

JPrevost:
#15 has some notion of roles, but I'd prefer to leave this open as it references a lot of other permissions-y things that are kinda super important.

JPrevost:
moved to https://mitlibraries.atlassian.net/browse/TI-6

Issue #9, opened 2017-07-11
Create (temporary but dedicated) MIT Pilot Auth credentials


JPrevost:
Register this app with MIT OpenID Pilot and share keys via LastPass

JPrevost:
moved to https://mitlibraries.atlassian.net/browse/TI-9

PR #15, opened 2017-07-25
Adds admin dashboards

JPrevost:
## Status
**READY**

#### What does this PR do?

* adds admin dashboard via `administrate` gem
* only Users with `admin = true` can access the dashboards
* An admin ui at this early stage will let us confirm the data models
  are indeed what we intend them to be and move more rapidly to staff
  feedback (not yet though!).
* `administrate` seems flexible enough to use for semi-custom admin
  interfaces without the hassle of weird DSLs.
* Most of what we will need for admin UI works out of the box. At this
  time I believe adding the rest won't be awful. You can remind me when
  I am wrong.

These are not intended to be complete admin interfaces. Instead, these are proposed
starting points for our admin interfaces.

Using credentials for MIT OpenID in your ENV (I know #9 is still open but for local dev you can use your own):

Confirm you have no admin rights:
```
heroku local:run rails console
u = User.find_by(email: 'youremail@mit.edu')
u.admin?
```

If you have admin, remove it for now.

```
u.admin = false
u.save
exit
```

If you have no account, no worries as one will be created with no admin rights in the next step if one doesn't exist.

```
heroku local
localhost:5000/users/auth/mit_oauth2
localhost:5000/admin
```

You should be on the "hi" page because we don't have flashes enabled or a template yet.

Back in the console, add your admin rights and then back on the site:
```
localhost:5000/admin
```

And you should have dashboard access and be able to CRUD all the models from the generated UI.

- #14 
- #13 (it's not much yet, but it's there)

#### Todo:
- [x] Tests
- [ ] Documentation

NO

YES

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ git_source(:github) do |repo_name|
   "https://github.com/#{repo_name}.git"
 end
 
+gem 'administrate'
 gem 'devise'
 gem 'omniauth-mit-oauth2'
 gem 'omniauth-oauth2'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -40,10 +40,23 @@ GEM
       tzinfo (~> 1.1)
     addressable (2.5.1)
       public_suffix (~> 2.0, >= 2.0.2)
+    administrate (0.8.1)
+      actionpack (>= 4.2, < 5.2)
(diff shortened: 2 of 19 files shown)


thatandromeda:
👍

PR #17, opened 2017-07-27
Styles

JPrevost:
## Status
**READY**

#### What does this PR do?

- Adds MIT Libraries styles (css and html header / footer components)
- includes sign in / sign out link
- includes link to admin if signed in and an admin

The sign in / sign out / admin links are primarily for convenience and not necessarily
intended to remain in that location after the UI is built out. (I mostly just keep forgetting how to sign in and needed a quick button before we build out the full UI).

This is not automatically deployed anywhere (sorry) so you can review in a local VM if you want to test things out yourself.  Once you have a local copy running, you'll immediately notice the MIT Libraries styles.

If you attempt to sign in though, you will need ENV setting for the MIT OpenID Pilot. I can assist with that if necessary.

- #16 

NO

NO


JPrevost:
assigning review to @frrrances because it's mostly styles but cc'ing @thatandromeda

frrrances:
lgtm. 👍

Issue #19, opened 2017-08-10
Initial user facing form


JPrevost:
Create a user facing form to collect metadata about a thesis.

The form:
- [ ] Users are prompted to sign before filling out the form
- [ ] New thesis is linked to `current_user`
- [ ] Title: string (required)
- [ ] Graduation Date: year / month from drop downs (required)
- [ ] Abstract: text
- [ ] Rights Assignment (dropdown from Rights): required
- [ ] Department (dropdown from Departments): required
- [ ] Degree (dropdown from Degrees): required

Post submission:
- [ ] Displays receipt when submitted

The following will be handled in separate Issues
- Uploading files will be handle in a separate Issue
- Advisors: multiple repeatable string field. required

PR #21, opened 2017-08-18
18 initial thesis submission form

JPrevost:
## Status
**READY**

#### What does this PR do?

Creates the initial user facing Thesis submission form.

Advisor validation is temporarily removed from the Thesis model to allow web form work to
proceed with the simpler elements before tackling this more complex form element. A new ticket #22  will handle Advisors in the web form and the validation and tests will be reenabled at that time.

- `bundle install` to get the new dependencies
- Locally, run the application with `heroku local` 
- access `/thesis/new`
- see the new (missing advisors and file uploads still) web form

- #19 

#### Todo:
- [x] Tests

NO

YES

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem 'omniauth-oauth2'
 gem 'puma'
 gem 'rails', '~> 5.1.2'
 gem 'sass-rails'
+gem 'simple_form'
 gem 'therubyracer', platforms: :ruby
 gem 'uglifier'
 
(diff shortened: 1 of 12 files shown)


thatandromeda:
Some things that should be considered as separate tickets:
* If I go to thesis/new and I'm not logged in, I get routed through OAuth and then dumped onto the root URL - I should get back to my original URL.
* I presume department and degree options are set by us - should submitters have any recourse if it turns out theirs are not available?

Anyway, please fix the codeclimate and then merge.

PR #28, opened 2017-12-01
Touchstone SAML

JPrevost:
## Status
**READY**

#### What does this PR do?

- changes auth from MIT Pilot to MIT Touchstone via in app SAML SP
- provides a dev mode auth option for PR builds and local work as
  registering a Touchstone SP for each PR and local environment is
  tedious
- also adds db migrate to release phase

https://mitlibraries.github.io/touchstone_saml.html

Login at https://library-thesis-dropbox-staging.mit.edu

#### Todo:
- [x] Tests
- [x] Documentation
- [x] Stakeholder approval

YES

YES

### DIFF
--- a/.adr-dir
+++ b/.adr-dir
@@ -1,0 +1,1 @@
+docs/architecture_decisions
--- a/README.md
+++ b/README.md
@@ -7,12 +7,56 @@
 This is a simple web app to collect metadata and files from a `User` and allow
 `Users` with the role `Admin` to download and edit the metadata.
 
+# Architecture Decision Records
+
+This repository contains Architecture Decision Records in the
+[docs/architecture-decisions directory](docs/architecture_decisions).
+
+[adr-tools](https://github.com/npryce/adr-tools) should allow easy creation of
+additional records with a standardized template.
+
 # Developer Notes
 
 When changing the db schema, please run `bundle exec annotate` to update the
 model and associated tests to reflect the changes in a nice convenient,
 consistent way.
 
+# Environment Variables
+
+## Authentication for Development ONLY
+
+`FAKE_AUTH_ENABLED` - set to `true`. WARNING: DO NOT DO THIS IN PRODUCTION! BAD!
+
+## Authentication for production
+
+For SAML authentication, you will need all of the following.
+
+[DLE Docs on SAML](https://mitlibraries.github.io/touchstone_saml.html)
+
+`IDP_METADATA_URL` - URL from which the IDP metadata can be obtained. This is
+  loaded at application start to ensure it remains up to date.
+
+`IDP_ENTITY_ID` - If `IDP_METADATA_URL` returns more than one IdP (like MIT
+  does) entry, this setting signifies which IdP to use.
+
+`IDP_SSO_URL` - the URL from the IdP metadata to use for authentication. I was
+  unable to extract this directly from the metadata with the ruby-saml tool
+  even though it for sure exists.
+
+`SP_ENTITY_ID` - unique identifier to this application,
+  ex: `https://example.com/shibboleth`
+
+`SP_PRIVATE_KEY` - Base64 strict encoded version of the SP Private Key.
+  note: Base64 is required due to multiline ENV being weird to deal with.
+
+`SP_CERTIFICATE` - Base64 strict encoded version of the SP Certificate.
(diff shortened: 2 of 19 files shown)


README.md
```diff
@@ -7,6 +7,14 @@
 This is a simple web app to collect metadata and files from a `User` and allow
 `Users` with the role `Admin` to download and edit the metadata.
 
+# Architecture Decision Records
+
+This repository contains Architecture Decision Records in the
+[docs/architecture-decisions directory](docs/architecture-decisions).
+
+[adr-tools](https://github.com/npryce/adr-tools) should allow easy creation of
+additional records with a standardize template.
```
thatandromeda: s/standardize/standardized/;

.adr-dir
```diff
@@ -0,0 +1 @@
+docs/architecture_decisions
```
thatandromeda: Woohoo, documentation!

app/controllers/application_controller.rb
```diff
@@ -1,5 +1,6 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
+  skip_before_action :verify_authenticity_token
```
thatandromeda: Wait, why?

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -1,9 +1,19 @@
 module Users
   class OmniauthCallbacksController < Devise::OmniauthCallbacksController
-    def mit_oauth2
+    def saml
       @user = User.from_omniauth(request.env['omniauth.auth'])
       sign_in_and_redirect @user, event: :authentication
       flash[:notice] = "Welcome #{@user.email}!"
     end
+
+    def developer
```
thatandromeda: This could stand to have a comment (and also the docs should be very clear about ensuring `FAKE_AUTH_ENABLED` is false in production).

thatandromeda: Super handy, though. I see why it's here.

thatandromeda:
Yay, progress!

This requires several environment variables to be set - can those variables and their desired settings be documented somewhere?

I am torn on the developer authentication. I see how this is super useful for testing purposes, but it is also very explicitly a thing that makes everyone an admin user, which is potentially a security-hole gun sitting around so that we can shoot ourselves with it. I would at the least like to see that function liberally festooned with documentation as to how _not_ to shoot ourselves with it, but if you have additional ideas for making it hard to trip over by accident that would be great too.

app/controllers/application_controller.rb
JPrevost: It's a form submission from the Touchstone system essentially so it does not have an authenticity token. It's from the docs.

app/controllers/users/omniauth_callbacks_controller.rb
JPrevost: Cool. I can doc that up. Readme for sure and a note here.

JPrevost:
Yeah. We need dev mode for PR builds as well as those URLs aren't registered with Touchstone but people will sometimes need to login to test stuff. I'll add some new docs and warnings around it as well as open a ticket to consider preventing it from running on the production heroku instances... but that gets a bit tricky as prod staging and pr builds all run with `production` as the environment flag. We can do better though with some focused thought I bet.

README.md
JPrevost: will fix

JPrevost:
@thatandromeda I've added notes on the ENV and added an initial warning. I will be opening a new ticket to further prevent enabling dev auth on staging / production servers.

JPrevost:
And the promised ticket to safety it up: https://mitlibraries.atlassian.net/browse/TI-23

README.md
```diff
@@ -13,14 +13,50 @@ This repository contains Architecture Decision Records in the
...
+
+## Authentication for Development ONLY
+
+`FAKE_AUTH_ENABLED` - set to `true`. WARNING: DO NOT DO THIS IN PRODUCTION! BAD!
+
+## Authentication for production
+
+For SAML authentication, you will need all of the following.
+
+[DLE Docs on SAML](https://mitlibraries.github.io/touchstone_saml.html)
```
thatandromeda: 🌈 for these existing.

PR #30, opened 2017-12-21
Logging and monitoring

JPrevost:
## Status
**READY**

#### What does this PR do?

- configures production application logging to be more 🌈 
- adds exception logging via rollbar
- adds apm via skylight

- https://mitlibraries.atlassian.net/browse/TI-12
- https://mitlibraries.atlassian.net/browse/TI-11
- https://mitlibraries.atlassian.net/browse/TI-30

NO

YES


Gemfile
```diff
@@ -8,11 +8,14 @@ end
 
 gem 'administrate'
 gem 'devise'
+gem 'lograge'
```
thatandromeda: 🌲 😡 

(sorry, it's just a funny gem name :)

config/environments/development.rb
```diff
@@ -48,6 +48,20 @@
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
 
+  if ENV['LOG_LIKE_PROD'].present?
+    config.lograge.enabled = true
+    config.lograge.custom_options = lambda do |event|
+      exceptions = %w(controller action format id)
```
thatandromeda: What does this actually *mean*?

thatandromeda: And is there some way that prod and log_like_prod can point to the *same* code? I'm worried that `LOG_LIKE_PROD` will get UNLIKE_PROD if we update the prod config and forget that this also needs to be. (If that's hard, the prod settings should at least have a comment warning us to update the dev config OR ELSE.)

config/initializers/rollbar.rb
```diff
@@ -0,0 +1,57 @@
+Rollbar.configure do |config|
+  # Without configuration, Rollbar is enabled in all environments.
+  # To disable in specific environments, set config.enabled=false.
+
+  config.access_token = ENV['ROLLBAR_ACCESS_TOKEN']
+
+  # Here we'll disable in 'test':
+  if Rails.env.test? || Rails.env.development?
```
thatandromeda: Do we want this, or do we want it configurable via env?

config/initializers/rollbar.rb
```diff
@@ -0,0 +1,57 @@
...
+    config.enabled = false
+  end
+
+  # By default, Rollbar will try to call the `current_user` controller method
```
thatandromeda: This sounds like we'll be putting PII in our logs. I would rather we not do that. Is there some compelling reason we must do that? Can we avoid doing that?

config/initializers/rollbar.rb
JPrevost: This is fine. In dev and test you'd just look at exceptions as they come up in your console.

config/initializers/rollbar.rb
JPrevost: Right now it indeed logs the email address associated with the user. I think it's useful to know which user is having a problem but agree we should consider what information we log along with the problem report. Would you feel more comfortable if we just logged the internal user id from the db table and not the email at all? Our staff could still look up the user, but it would need to be intentional. (We could really use guidelines).

config/environments/development.rb
JPrevost: @gravesm recently unified the log configs. You can still switch them off if you need the default behavior, but it does move them to a single initializer. I'll do the same here.

The exceptions here effectively remove some duplicate data from the logs to make them a bit more readable.

thatandromeda: @gravesm 🍺 🎆

config/initializers/rollbar.rb
thatandromeda: I would feel better about the user ID, yes. (Although if "internal user ID" = "MIT ID" we have to be very careful not to log any other personal information or I think we fall afoul of MIT's info protection guidelines.)

JPrevost:
@thatandromeda I've added two commits to address your feedback. One to move the lograge configuration to a shared initializer and the second to ensure not PII is sent to the exception monitor.

thatandromeda:
👍

PR #32, opened 2018-01-03
23 desuckify security

thatandromeda:
## Status
**READY**

#### What does this PR do?
We need to ensure that we don't accidentally enable fake auth in production. This checks on the Heroku app name as well as the `FAKE_AUTH_ENABLED` env variable.

- https://mitlibraries.atlassian.net/browse/TI-23

#### Todo:
- [ ] Tests
- [ ] Documentation
- [ ] Stakeholder approval

NO

YES (for some reason dotenv wasn't already there)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,6 +38,7 @@ group :development do
 end
 
 group :test do
+  gem 'climate_control'
   gem 'coveralls', require: false
   gem 'minitest-reporters'
 end
--- a/app.json
+++ b/app.json
@@ -1,0 +1,53 @@
+{
+  "name": "thing",
+  "scripts": {},
+  "env": {
+    "HEROKU_APP_NAME": {
+      "required": true
+    },
+    "LANG": {
+      "required": true
+    },
+    "PAPERTRAIL_API_TOKEN": {
+      "required": true
+    },
+    "PRIVATE_KEY": {
+      "required": true
+    },
+    "RACK_ENV": {
+      "required": true
+    },
+    "RAILS_ENV": {
+      "required": true
+    },
+    "RAILS_LOG_TO_STDOUT": {
+      "required": true
+    },
+    "RAILS_SERVE_STATIC_FILES": {
+      "required": true
+    },
+    "SECRET_KEY_BASE": {
+      "required": true
+    },
+    "SKYLIGHT_AUTHENTICATION": {
+      "required": true
+    },
+  },
+  "formation": {
+    "web": {
+      "quantity": 1
+    },
+    "worker": {
+      "quantity": 1
+    }
+  },
+  "addons": [
+    "papertrail",
+    "heroku-postgresql"
+  ],
+  "buildpacks": [
+    {
+      "url": "heroku/ruby"
+    }
+  ]
+}
--- a/app/models/concerns/fake_auth_config.rb
+++ b/app/models/concerns/fake_auth_config.rb
@@ -1,0 +1,26 @@
+module FakeAuthConfig
+  # Used in an initializer to determine if the application is configured and
+  # allowed to use fake authentication.
+  def fake_auth_status
+    return true if fake_auth_enabled? && app_name_pattern_match?
+    false
+  end
+
+  private
+
+  def fake_auth_enabled?
+    ENV['FAKE_AUTH_ENABLED'] == 'true'
+  end
+
+  # Checks to make sure the application is not the staging or production
+  # instance via a heroku ENV.
+  # In development env we always return `true` to avoid having to set a fake app
+  # name to match the pr build name patterns.
+  # In test env we require setting a fake app name to allow for testing of the
+  # pattern.
+  def app_name_pattern_match?
+    return true if Rails.env.development?
+    review_app_pattern = /^library-thesis-dropbox-staging-pr-[\d]+$/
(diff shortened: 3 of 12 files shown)


thatandromeda:
@JPrevost This isn't done but I had some questions and this seemed like the easiest way to ask them.

* Does using `Rails.configuration.fake_auth_enabled` make sense? Using `Rails.configuration` isn't a pattern we use elsewhere so I wonder if it feels too weird.
* If not that, what? I thought about making it a method in ApplicationController or ApplicationHelper, but we need to check it from OmniauthCallbacksController (does not inherit from ApplicationController) and also `config/initializers/devise.rb`, which I think means I can't make it later than application config.
* That said I don't know if application config is even testable (how would i check to see that env had been properly translated into config without reinitializing config, and how can I reinitialize in the middle of running tests?) A helper method would be way easier to test but I would not be able to include it in the Devise initialization (would I?)

Anyway, I'm hoping you know some magic Railsy way around all this.

app.json
```diff
@@ -0,0 +1,83 @@
+{
+  "name": "thing",
+  "scripts": {},
+  "env": {
+    "CERTIFICATE": {
```
JPrevost: We shouldn't move the certs or IDP, URN, and SP stuff to pr apps. It won't work as they won't be registered in touchstone.

app.json
```diff
@@ -0,0 +1,83 @@
...
+      "required": true
+    },
+    "HEROKU_APP_NAME": {
```
JPrevost: Hmmm. I thought `required: true` copied the value from the staging app which would not be desirable. Maybe I don't get it.

Gemfile
```diff
@@ -25,6 +25,7 @@ group :development, :test do
   gem 'capybara'
   gem 'selenium-webdriver'
   gem 'sqlite3'
+  gem 'dotenv'
```
JPrevost: `heroku local` already loads stuff from `.env`. What are we gaining with `dotenv` gem? I'm not opposed to this if you find it better, but we will need to be careful as it can get weird with test envs if we aren't careful.

config/application.rb
```diff
@@ -14,5 +14,14 @@ class Application < Rails::Application
     # Settings in config/environments/* take precedence over those specified here.
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
+
+    Dotenv.load  # We need this to perform the checks below.
```
JPrevost: Oh. Hmmm. I'm not sure I love this. I'll get back to you but I feel like loading ENV outside of the normal way is asking for trouble longterm. The `dotenv` gem usually just loads without this anyway so it's possible this is not necessary?

config/application.rb
```diff
@@ -14,5 +14,14 @@ class Application < Rails::Application
...
+    review_app_pattern = /^library-thesis-dropbox-(staging-)?pr-[\d]+$/
```
JPrevost: All PR builds should be based on the staging app so the parens should be safe to remove.

JPrevost:
@thatandromeda I added a commit that moves the configuration to an initializer. I'm sorry I didn't just provide feedback to suggest doing that, but I wasn't entirely sure whether it would work out and honestly was not as straight forward as I initially assumed it would be. I moved to a module to allow for easier testing and am using a gem I've been meaning to try out to modify ENV in tests in a clean way. Thoughts are very welcome and again sorry for providing code rather than coherent thoughts on your PR :)

JPrevost:
@thatandromeda also, I removed dotenv as I couldn't get it to reliably work without calling `Dotenv.load` explicitly which I think would have errored fatally in production unless we included the gem outside of the development group...which I'm not sure I love. Maybe who care? I haven't decided. Either way, `heroku local` currently works as expected. I do want to discuss (soonish?) using `dotenv` vs `heroku local` as our standard for rails apps as not everything will target heroku but there is no good reason to not have everything using `.env` files in development.

app.json
thatandromeda: HEROKU_APP_NAME follows a different pattern, which is weird: https://devcenter.heroku.com/articles/github-integration-review-apps

app.json
thatandromeda: 👍 Removing them.

app/models/concerns/fake_auth_config.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  private
+
+  def fake_auth_enabled?
+    ENV['FAKE_AUTH_ENABLED'] == 'true'
+  end
+
+  # Checks to make sure the application is not the staging or production
+  # instance via a heroku ENV.
+  # In development env we always return `true` to avoid having to set a fake app
```
thatandromeda: 👍

config/initializers/0_fake_auth_configuration.rb
```diff
@@ -0,0 +1,5 @@
+# This configuration needs to load before the devise initializer so this value
```
thatandromeda: 👍 Good thing to document in a comment.

app.json
JPrevost: Ah, weird, but okay :)

thatandromeda:
@JPrevost I started the conversation in code rather than English; it's perfectly legit for you to continue it that way.

Your changes look 🌈 . I've rebased onto updated master and squashed - this is ready for review.

PR #36, opened 2018-01-19
Front page is splash for anon users, thesis submission for logged-in

thatandromeda:
## Status
**READY**

#### What does this PR do?
Root path is a splash page to anonymous users (pending TI-25), thesis submission page for logged-in users.

Fire up the app. Lo, it is the splash page. Log in. Lo, it is now the thesis submission form.

This was originally intended as just TI-24, but adding the test for redirect-after-login behavior turned out to be logically entailed by that work, so this will also close TI-26.

- https://mitlibraries.atlassian.net/browse/TI-26
- https://mitlibraries.atlassian.net/browse/TI-24

#### Todo:
- [x] Tests
- [ ] Documentation
- [ ] Stakeholder approval

NO

NO

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -21,5 +21,9 @@ Rails.application.routes.draw do
     delete 'sign_out', to: 'devise/sessions#destroy', as: :destroy_user_session
   end
 
+  authenticated do
+    root :to => 'thesis#new'
+  end
+
   root to: 'static#index'
 end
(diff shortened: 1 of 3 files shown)


thatandromeda:
It seemed to me like a single-serving site, and also that's the literal interpretation of the comment I have in the ticket following our earlier discussion. But I don't have strong feelings about it.

Are we anticipating the splash page might have content people actually want to reread, like an about page?

JPrevost:
Yeah, I have no idea. If you feel okay about this with the added context I put in I an get behind it (we can always change later if stakeholders don't like this).

thatandromeda:
If we decide the splash page has info people might actually need, we can also adapt it so that the logged-in view is submisison form + splash content, I guess. Pull in the partial.

PR #37, opened 2018-01-19
Adds initial roles

JPrevost:
## Status
**READY**

#### What does this PR do?

Adds initial roles and tests to confirm expectations.

https://github.com/CanCanCommunity/cancancan

The PR builds are broken (ticket is open) but if you look at this locally you can create a couple users and give one a thesis and confirm the other can't see it.

- https://mitlibraries.atlassian.net/browse/TI-6

#### Todo:
- [x] Tests
- [x] Documentation
- [ ] Stakeholder approval

NO

YES

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -1,0 +1,22 @@
+class Ability
+  include CanCan::Ability
+
+  def initialize(user)
+    # Define abilities for the passed in user here.
+    # See the documentation for details:
+    # https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities
+
+    if user.present?
+      # Any user can create a new Thesis
+      can :create, Thesis
+
+      # Only the Thesis owner can view their Thesis
+      can :read, Thesis, user_id: user.id
+
+      # Admin users can do everything for all models
+      if user.admin?
+        can :manage, :all
+      end
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


test/controllers/thesis_controller_test.rb
```diff
@@ -58,4 +58,78 @@ class ThesisControllerTest < ActionDispatch::IntegrationTest
...
+    assert_response :redirect
+  end
+
+  test 'admin users can view another user thesis' do
+    sign_in users(:admin)
+    get "/thesis/#{theses(:one).id}"
+    assert_response :success
+  end
+
+  # Tests below this note are to protect us from our future selves.
```
thatandromeda: 👍

app/models/ability.rb
```diff
@@ -0,0 +1,20 @@
...
+  def initialize(user)
+    # Define abilities for the passed in user here.
+    # See the documentation for details:
+    # https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities
+
+    # Any user can create a new Thesis
+    can :create, Thesis
+
+    # Only the Thesis owner can view their Thesis
+    can :read, Thesis, user_id: user.id
```
thatandromeda: The docs wrap tests of user attributes in an `if user.present?`, but this seems to work even if we don't do that. I guess it's cool? (ISTR rails and django handle anonymous users differently but I don't follow the specifics.)

JPrevost: oh. right. Let me dig in on that a bit more as now I'm not sure...hmmm.

JPrevost:
I've updated abilities.rb to match the documentation. Thanks for the catch @thatandromeda

PR #38, opened 2018-01-19
valid json ftw

JPrevost:
## Status
**READY**

#### What does this PR do?

Apparently valid JSON is a good thing.

If the PR build works, it work!

- https://mitlibraries.atlassian.net/browse/TI-36

NO

NO

### DIFF
--- a/app.json
+++ b/app.json
@@ -27,7 +27,7 @@
     "ROLLBAR_ENV": "pr",
     "SECRET_KEY_BASE": {
       "required": true
-    },
+    }
   },
   "formation": {
     "web": {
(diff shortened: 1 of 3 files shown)


thatandromeda:
The PR build does not work :(

thatandromeda:
I think it's failing because it lacks IDP metadata. And IIRC there are valid reasons not to have that on PR builds, but there's a devise config thing that dies without it.

thatandromeda:
Anyway, if the goal of this PR is to have valid JSON I'm good with it, but if the goal is to fix PR builds (which I think is the better goal), there is more work to be done.

JPrevost:
This PR is flagged as In Development ;) I know it's broke but went to lunch when it was clear it was not a quick fix lol.

JPrevost:
@thatandromeda should be working now after updating the match pattern.

PR #43, opened 2018-01-30
Let users filter thesis processing page by status

thatandromeda:
## Status
**READY**

#### What does this PR do?
Adds filter-by-status options to the thesis processing queue; tests them.

Go to `/process` on the PR build. Observe the view-by buttons. Click them.

- https://mitlibraries.atlassian.net/browse/TI-38

#### Todo:
- [x] Tests
- [ ] Documentation
- [ ] Stakeholder approval

NO

NO

### DIFF
--- a/test/controllers/thesis_controller_test.rb
+++ b/test/controllers/thesis_controller_test.rb
@@ -264,6 +264,117 @@ class ThesisControllerTest < ActionDispatch::IntegrationTest
     assert_select "div[data-id=thesis_#{thesis.id}] form", count: 0
   end
 
+  # ~~~~~~~~~~~~~~~~~~~~ filters on the processing queue ~~~~~~~~~~~~~~~~~~~~~~
+  test 'queryset when no status is given' do
+    sign_in users(:admin)
+    get process_path
+
+    # We want to make sure that this queryset returned by the controller is the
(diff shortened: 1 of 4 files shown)


test/controllers/thesis_controller_test.rb
```diff
@@ -264,6 +264,117 @@ class ThesisControllerTest < ActionDispatch::IntegrationTest
     assert_select "div[data-id=thesis_#{thesis.id}] form", count: 0
   end
 
+  # ~~~~~~~~~~~~~~~~~~~~ filters on the processing queue ~~~~~~~~~~~~~~~~~~~~~~
+  test 'queryset when no status is given' do
+    sign_in users(:admin)
+    get process_path
+
+    # We want to make sure that this queryset returned by the controller is the
```
JPrevost: I feel a bit like this is sliding a bit towards testing that ActiveRecord works... which is out of scope of our app concerns. I'm not suggesting we remove them as they have already been written, but I think the second set of tests is adequately 🌈

thatandromeda: I think it's a little more subtle than that in that it's testing that I actually wrote the queries that I intended to write, and they are wired up to parameters in the way that I intended to wire them up. (I hear you on not testing rails internals, though.)

PR #44, opened 2018-01-30
34 refactor has and belongs to many

JPrevost:
## Status
**READY**

#### What does this PR do?

- drops advisor remnants as they are not currently needed in app (this is included as it would have needed refactoring if not dropped)
- refactors Theses/Degree relationship to use a join table matching best practices
- refactors Theses/Departments relationship to use a join table matching best practices

http://blog.flatironschool.com/why-you-dont-need-has-and-belongs-to-many/

- https://mitlibraries.atlassian.net/browse/TI-34

#### Todo:
- [x] Tests
- [x] Documentation

YES

NO


app/models/degree_thesis.rb
```diff
@@ -0,0 +1,12 @@
+# == Schema Information
+#
+# Table name: degree_theses
```
thatandromeda: Why is one of these plural and one of them singular?

app/models/department.rb
```diff
@@ -9,7 +9,8 @@
 #
 
 class Department < ApplicationRecord
-  has_and_belongs_to_many :theses
+  has_many :department_theses
```
thatandromeda: (same singular/plural question)

thatandromeda:
I would like to be convinced we're not confusing future-us with the singular-vs-plural name thing (and also there's a rebase). Other than that, looks 🌈 .

app/models/department.rb
JPrevost: I oddly ran into problems unless I did it this way. I'm confident it was solvable with different variants, but I landed on not really caring enough to figure out why I couldn't use all plural or all singular without jumping through hoops. Sorry I lazied it :/

thatandromeda: Coolio. Can you leave a comment so we're not confused in future? Then go ahead with the rebase/squash/merge thing.

PR #61, opened 2018-02-16
Use permalinks, not expiring service URLs

thatandromeda:
## Status
**READY**

#### What does this PR do?
`file.service_url` expires; we should use a nonexpiring link to thesis content. I'm pretty sure this is the Rails-approved, All We Need To Do thing.

Compare file URL structures on staging and the PR build - they look different.

Save a staging build and PR build URL for >5 minutes and then go back to them - PR build one should still work, staging will have expired.

- https://mitlibraries.atlassian.net/browse/TI-55

#### Todo:
- [ ] Tests
- [ ] Documentation
- [ ] Stakeholder approval

NO

NO

### DIFF
--- a/app/views/thesis/process_theses.html.erb
+++ b/app/views/thesis/process_theses.html.erb
@@ -52,7 +52,7 @@
           Files:  <% if thesis.files.attached? %>
                     <% thesis.files.each do |file| %>
                       <br />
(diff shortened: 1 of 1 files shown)


thatandromeda:
For the record: I have now tested my claim about expiring staging vs non-expiring PR build links (I was right). I have no idea how to automate that testing.

PR #63, opened 2018-02-23
Add name field to Users

thatandromeda:
## Status
**Ready**

**NOTE THAT DEPLOYMENT WILL REQUIRE SETTING NEW ENVIRONMENT VARIABLES**. These are documented in the README.

#### What does this PR do?
Adds names to the user model. Grabs those names from Shib.

We don't actually _require_ names at the database level because we have existing users without names, but we do validate it for new users.

Oh, and fixes the fake auth bug (now that I understand how auth works) and a bug whereby non-admin thesis processors didn't actually get to see the thesis processing link (whoops, since I'm an admin I never saw that before).

This branch is up _on staging_ (because it, unlike test servers, can talk to Touchstone).

1) delete your kerb user on staging
2) go to staging _with the MIT URL_
3) sign in with touchstone
4) note that your newly created user has a name

You'll probably have to perform some of these verification steps via the shell.

- https://mitlibraries.atlassian.net/browse/TI-42
- https://mitlibraries.atlassian.net/browse/TI-48

#### Todo:
- [x] Tests
- [ ] Documentation
- [x] Stakeholder approval

YES

NO


README.md
```diff
@@ -99,7 +99,9 @@ For SAML authentication, you will need all of the following.
 
 `URN_EMAIL` - URN to extract from SAML response. For MIT, `urn:oid:0.9.2342.19200300.100.1.3` for testshib `urn:oid:1.3.6.1.4.1.5923.1.1.1.6` is close enough for testing.
 
-`URN_NAME` - `urn:oid:2.16.840.1.113730.3.1.241` for MIT, `urn:oid:2.5.4.3` for testshib
+`URN_GIVEN_NAME` - `urn:oid:2.5.4.42` for MIT
+`URN_SN` - `urn:oid:2.5.4.4` for MIT
```
JPrevost: `URN_SURNAME` may be clearer

README.md
```diff
@@ -99,7 +99,9 @@ For SAML authentication, you will need all of the following.
...
+These have not been tried with testshib but there's a chance they're the same.
```
JPrevost: I tested in testshib and these work fine. We can update the note to reflect they are the same for both.

app/models/user.rb
```diff
@@ -28,9 +32,25 @@ class User < ApplicationRecord
...
+    User.where(uid: uid).first_or_create do |user|
       user.email = auth.info.email
+      user.given_name = auth.info.given_name
+      user.surname = auth.info.surname
     end
   end
+
+  # Definitely for sure wrong for some people. But staff need name info to
+  # process theses and we have no way of knowing which users would prefer
+  # something else.
```
JPrevost: Would it be better to use the displayName attribute we can get from Touchstone that _may_ address some of these concerns? I feel like what we have here is fine, but since you commented it's probably wrong sometimes maybe we should consider using the best data we have?

app/views/layouts/_site_header.html.erb
```diff
@@ -23,9 +23,13 @@
     <div class="wrap-local-nav">
       <div class="wrap-bar">
         <nav class="local-nav" aria-label="Main menu">
-          <% if user_signed_in? && current_user.admin? %>
-            <%= link_to("Process submissions", process_path, class: 'button-secondary') %>
-            <%= link_to("Admin", admin_root_path, class: 'button-secondary') %>
+          <% if user_signed_in? %>
+            <% if can? :process_theses, Thesis %>
```
JPrevost: Ah... I was trying to write up this bug but wasn't sure what was going on enough to log it even. I had a local note to figure it out and open a bug. 💯

README.md
thatandromeda: I agree. I went with SN because that's what's in the Shib docs 🙈 , but `URN_SURNAME` is better. I've noted the disparity in the readme.

README.md
thatandromeda: 👍

app/models/user.rb
thatandromeda: If the whole point is that we're sorting on last name, we need something normalized to last name. If we find ourselves actually showing names to users, maybe we should get displayName. I'll add the comment.

PR #66, opened 2018-02-27
Email receipt of Thesis submissions

JPrevost:
## Status
**READY**

#### What does this PR do?

- sends receipt emails
- emails are disabled in PR builds by default (but are _enabled_ on this PR intentionally)
- delivers emails as delayed job so a worker process is required

- sign in with a real email address
- submit a thesis
- check your email

- https://mitlibraries.atlassian.net/browse/TI-28

#### Todo:
- [x] Tests
- [x] Documentation
- [ ] Stakeholder approval

YES

YES


README.md
```diff
@@ -51,6 +51,24 @@ of variables:
 In addition, you will need to ensure the bucket CORS `AllowedOrigin` settings
 are configured to allow for the domain this app runs at.
 
+## Sending Receipt Email in Production
+
+`SMTP_ADDRESS`
```
thatandromeda: Do we need these 5 set in every environment or just on prod?

If we don't need all 5 everywhere, what needs to be set on local?

What needs to be set on heroku when `DISABLE_ALL_EMAIL`?

README.md
```diff
@@ -51,6 +51,24 @@ of variables:
...
+`SMTP_USER`
+`THESIS_ADMIN_EMAIL` - used for both `from` and `cc` of receipt emails.
+
+In development, emails are written to a file in `tmp`. In testing, they are
+stored in memory. You still need the `THESIS_ADMIN_EMAIL` set for the tmp file
+to be written without errors.
+
+On staging, the default is disabled email. Unset `DISABLE_ALL_EMAIL` if you
+have a reason to turn them on. Due to the potential of unwanted emails being
+sent when `FAKE_AUTH_ENABLED` is enabled (like on PR builds), it's best to
```
thatandromeda: Should we make the fake auth setting override the disable all email setting, just in case?

app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,4 @@
 class ApplicationMailer < ActionMailer::Base
-  default from: 'from@example.com'
+  default from: ENV['THESIS_ADMIN_EMAIL']
```
thatandromeda: Do we want to test for the presence of this key and have some kind of default (libdev-cs?) Or is it actually desirable for something to crash if this is unset?

app/views/receipt_mailer/receipt_email.html.erb
```diff
@@ -0,0 +1,22 @@
+<p>Hello,</p>
```
thatandromeda: Do we have stakeholder signoff?

app/views/receipt_mailer/receipt_email.html.erb
```diff
@@ -0,0 +1,22 @@
...
+<p>You do not need to reply to this message.</p>
+
+<p>Theses will be added to DSpace after degrees have been awarded and the thesis
+has been fully processed by the MIT Libraries (typically 3-6 months after the
+degree date).</p>
+
+<p>If you require further assistance, please send an email to
+etheses-admin@mit.edu.</p>
+
+<p>Thank you,<br />
```
thatandromeda: what, no "and congratulations"? ;)

app/views/receipt_mailer/receipt_email.html.erb
```diff
@@ -0,0 +1,22 @@
+<p>Hello,</p>
+
+<p>You are receiving this message to confirm the successful submission of the
+electronic version of your <%= @thesis.graduation_month %> <%=
+@thesis.graduation_year %> thesis "<%= @thesis.title %>" to the MIT Libraries.
+</p>
+
+<p>You do not need to reply to this message.</p>
+
+<p>Theses will be added to DSpace after degrees have been awarded and the thesis
```
thatandromeda: should dspace be a link?

What happens to people who want a text version of their email? (Does ruby autogenerate multipart with a text version from the html?)

app/views/receipt_mailer/receipt_email.html.erb
```diff
@@ -0,0 +1,22 @@
...
+has been fully processed by the MIT Libraries (typically 3-6 months after the
+degree date).</p>
+
+<p>If you require further assistance, please send an email to
+etheses-admin@mit.edu.</p>
```
thatandromeda: Should this be a mailto link?

app/views/receipt_mailer/receipt_email.html.erb
```diff
@@ -0,0 +1,22 @@
...
+
+<p>Thank you,<br />
+MIT Libraries</p>
+
+<p>We would like your feedback!<br />
+Please take a moment to fill out a short survey about this service:<br />
+<a href="https://goo.gl/QwE7ay">https://goo.gl/QwE7ay</a></p>
```
thatandromeda: Do we have any way whatsoever to monitor for linkrot?

config/environments/development.rb
```diff
@@ -31,9 +31,10 @@
   config.active_storage.service = :local
 
   # Don't care if the mailer can't send.
```
thatandromeda: Comment should be changed to match new behavior

config/environments/production.rb
```diff
@@ -69,6 +69,16 @@
   # Set this to true and configure the email server for immediate delivery to raise delivery errors.
   # config.action_mailer.raise_delivery_errors = false
 
+  config.action_mailer.delivery_method = :smtp
+  config.action_mailer.smtp_settings = {
+    user_name: ENV['SMTP_USER'],
```
thatandromeda: Are there values for these settings in lastpass?

test/fixtures/receipt_mailer/receipt_email
```diff
@@ -0,0 +1,34 @@
...
+    <style>
+      /* Email styles need to be inline */
+    </style>
+  </head>
+
+  <body>
+    <p>Hello,</p>
+
+<p>You are receiving this message to confirm the successful submission of the
+electronic version of your July 2017 thesis "MyString" to the MIT Libraries.
```
thatandromeda: Do we have a requirement that graduation date can't be too far in the past? Will this test eventually fail on that basis?

test/mailers/previews/receipt_mailer_preview.rb
```diff
@@ -0,0 +1,6 @@
+# Preview all emails at http://localhost:3000/rails/mailers/receipt_mailer
```
thatandromeda: all emails in the queue? or ever?

test/mailers/receipt_mailer_test.rb
```diff
@@ -0,0 +1,32 @@
+require 'test_helper'
+
+class ReceiptMailerTest < ActionMailer::TestCase
+  def test_receipt_email
+    thesis = theses(:one)
+    email = ReceiptMailer.receipt_email(thesis)
+
+    # Send the email, then test that it got queued
+    assert_emails 1 do
+      email.deliver_now
```
thatandromeda: Do we clean up the email-delivery location after the test?

README.md
JPrevost: If you want to send email, you need to configure the email with these settings. Dev and test never actually send email (even if you set those). Dev writes to a file and test writes to an in memory location (an array). Can you re-read this section and let me know if you have specific suggestions of what to change as I thought I covered that.

README.md
JPrevost: I don't think so. On PR builds you'd have to make several intentional choices to send email.

app/mailers/application_mailer.rb
JPrevost: I'd lean towards a runtime error if the app isn't configured usefully myself, but I'd listen to a different argument.

app/views/receipt_mailer/receipt_email.html.erb
JPrevost: This is the stakeholder supplied email text with the change of not using the first name like that suggested for "reasons".

app/views/receipt_mailer/receipt_email.html.erb
JPrevost: Text email is not default. We can add it easily. I'll open a ticket though as I'd like to discuss pros and cons outside the context of this PR.

app/views/receipt_mailer/receipt_email.html.erb
JPrevost: ha. Not my words sadly.

app/views/receipt_mailer/receipt_email.html.erb
JPrevost: no

config/environments/development.rb
JPrevost: will fix 👍

config/environments/production.rb
JPrevost: They are common for all of the rails apps that send email... so sort of if you look at bento instead of the non-existent thing settings. Thing settings will be added as we approach completion before we consider production "production".

test/mailers/previews/receipt_mailer_preview.rb
JPrevost: It generates a sample receipt email based on the first thesis in the system on your dev environment. If you have not, it'll be blank.

test/mailers/receipt_mailer_test.rb
JPrevost: No, testing of emails uses an in memory array.

test/fixtures/receipt_mailer/receipt_email
JPrevost: Hmmm. Good question. The Thesis fixture file that is loaded to compare against this email fixture would become invalid in that case... so this test isn't the problem so much as the upstream fixture itself. I'll open a ticket to review that concern as I think it is valid but outside the scope of the email test itself.

JPrevost: https://mitlibraries.atlassian.net/browse/TI-66

config/environments/production.rb
JPrevost: https://mitlibraries.atlassian.net/browse/TI-67

app/views/receipt_mailer/receipt_email.html.erb
JPrevost: https://mitlibraries.atlassian.net/browse/TI-68

config/environments/development.rb
JPrevost: outdated comment removed. 🌈

JPrevost:
@thatandromeda I've tried to address comments inline. As there are a lot of them... let me know if a quick chat would be better to discuss if anything remains necessary to change on this PR in terms of docs or extra tickets to open or anything else. Thx!

app/mailers/application_mailer.rb
thatandromeda: 👍 , just want to make sure this is a specific choice (and also that we have documented that this must always be set)

README.md
thatandromeda: The confusion I'm having is that (I think) four of these only need to be set if you're using SMTP, but one of them needs to be set even if you're not because you'll end up with a runtime error. (And also the thesis_admin_email setting is documented in a section whose header includes "...in production", but you have to also set it in other environments.)

I think I would like more clarity as to when you set what. `Sending Receipt Email in Production` vs `Handling email in other environments` (which might then have subsections `to test email` and `to disable email`?)

There's a lot of moving parts here and I want to be certain I know how to move them to accomplish different scenarios.

app/views/receipt_mailer/receipt_email.html.erb
thatandromeda: 👍

app/views/receipt_mailer/receipt_email.html.erb
thatandromeda: 🙈

app/views/receipt_mailer/receipt_email.html.erb
thatandromeda: 👍

config/environments/production.rb
thatandromeda: 👍

test/mailers/receipt_mailer_test.rb
thatandromeda: Sweet.

README.md
JPrevost: Thx for the clarification on the cover. I think I can make words that explain it better. 👍

JPrevost: I've put in a very small change to clarify the dev requirement for email in the dev env section. I also opened https://mitlibraries.atlassian.net/browse/TI-69 as I think there is room for significant improvement in grouping the information which ultimately led to this concern (not being consistent in how we are documenting settings / features).

PR #74, opened 2018-03-02
add link to processing from admin and file downloads from show

JPrevost:
## Status
**READY**

#### What does this PR do?

- adds link to process theses in admin navigation
- adds link to download files from admin thesis show pages

- in the admin interface, the top link in the navigation should always be to the processing page
- in the admin thesis view page, links to download files should now exist if the thesis has files attached

- https://mitlibraries.atlassian.net/browse/TI-63
- and another one but Jira is down so I am not sure the number

#### Todo:
- [ ] Tests
- [ ] Documentation
- [ ] Stakeholder approval

NO

NO

### DIFF
--- a/app/fields/file_field.rb
+++ b/app/fields/file_field.rb
@@ -1,0 +1,7 @@
+require 'administrate/field/base'
+
+class FileField < Administrate::Field::Base
+  def files
+    data.record.files
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -11,7 +11,7 @@ Rails.application.routes.draw do
   end
 
   resources :thesis, only: [:new, :create, :show]
-  get 'process', to: 'thesis#process_theses'
+  get 'process', to: 'thesis#process_theses', as: 'process'
   get 'process/:status', to: 'thesis#process_theses'
   post 'done/:id', to: 'thesis#mark_downloaded',
                    id: /[0-9]+/,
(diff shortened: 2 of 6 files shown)


app/fields/file_field.rb
```diff
@@ -0,0 +1,7 @@
+require 'administrate/field/base'
+
+class FileField < Administrate::Field::Base
+  def file_url
```
thatandromeda: `file_url` seems like a confusing name here for what looks like a possibly-plural set of things that maybe aren't URLs (as suggested by the `files` in the next line, and the `each` in the `_show` partial below). What sort(s) of thing(s) is this function actually returning?

config/routes.rb
```diff
@@ -11,7 +11,7 @@
   end
 
   resources :thesis, only: [:new, :create, :show]
-  get 'process', to: 'thesis#process_theses'
+  get 'process', to: 'thesis#process_theses', as: 'process'
```
thatandromeda: 👍

app/fields/file_field.rb
JPrevost: Good catch. My initial plan was to return just, er, a `file_url` but the administrate thing didn't seem to work the way my brain was working so I just changed to what worked easily... and didn't change the name. I'll change it to `files` maybe as it returns all of the files associated with a single Thesis.

PR #92, opened 2018-03-16
Restrict valid months to February, June, September

thatandromeda:
This requires changing a lot of files because the test fixtures
had different months and therefore this validity check caused
numerous tests to fail, but it isn't actually a very big change.

## Status
**READY**

On submission form on PR build, you will see that only February, June, and September are available in the graduation month drop-down. If you try to create a thesis with a different graduation month (e.g. in the shell), it will fail.

- https://mitlibraries.atlassian.net/browse/TI-85

#### Todo:
- [x] Tests
- [ ] Documentation
- [x] Stakeholder approval

NO

NO

### DIFF
--- a/app/views/thesis/new.html.erb
+++ b/app/views/thesis/new.html.erb
@@ -31,7 +31,7 @@
       <div class='group-inline'>
 
       <%= f.input :graduation_month, label: 'Month', as: :select,
-                           collection: Date::MONTHNAMES.compact,
+                           collection: Thesis::VALID_MONTHS,
                            wrapper_html: { class: 'field-wrap' },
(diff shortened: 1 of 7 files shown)


app/views/thesis/new.html.erb
```diff
@@ -31,7 +31,7 @@
       <div class='group-inline'>
 
       <%= f.input :graduation_month, label: 'Month', as: :select,
-                           collection: Date::MONTHNAMES.compact,
+                           collection: Thesis::VALID_MONTHS,
```
JPrevost: Nice. I hadn't looked at how that was originally coded and wasn't sure if this was gonna bit a pain or not. 👍

PR #94, opened 2018-03-19
allow access to splash page after login AND update breadcrumbs

JPrevost:
## Status
**READY**

#### What does this PR do?

Allows access to splash page after login

Previously we overwrote routes to accomplish this, but we need a way
for users to see the splash page information after login so we are
handling with conditional after login redirects using Devise features.

https://mitlibraries.atlassian.net/browse/TI-74

Updates breadcrumbs

https://mitlibraries.atlassian.net/browse/TI-79

NO

NO


app/controllers/application_controller.rb
```diff
@@ -4,4 +4,12 @@ class ApplicationController < ActionController::Base
   def new_session_path(_scope)
     root_path
   end
+
```
thatandromeda: Oh hey, that's a smart default.

app/views/layouts/_breadcrumbs.erb
```diff
@@ -1,7 +1,9 @@
 <div class="wrap-outer-breadcrumb layout-band">
   <div class="wrap-breadcrumb" role="navigation" aria-label="breadcrumbs">
     <div class="breadcrumbs">
-      <a href="https://libraries.mit.edu" title="MIT Libraries">Home</a>  &raquo;  <a href="/">Thesis Submission</a>
+      <%= link_to('Home', root_path) %>
```
thatandromeda: Should the filename be `_breadcrumbs.html.erb`?

app/views/layouts/_breadcrumbs.erb
```diff
@@ -1,7 +1,9 @@
...
+      &raquo;
```
thatandromeda: This breadcrumb file appears to be displaying a navigation path but isn't actually conditional on the user's location. Should it be?

test/integration/authentication_test.rb
```diff
@@ -40,9 +40,15 @@ def silence_omniauth
     assert_equal(usercount + 1, User.count)
   end
 
-  test 'redirect to root path after login' do
+  test 'redirect to root path after login for normal users' do
```
thatandromeda: We've scrapped the test that anon users see the splash page - did we want to scrap that?

app/views/layouts/_breadcrumbs.erb
JPrevost: yes. I "fixed" it from `_breadcrumbs.html` wrongly. Sigh.

app/views/layouts/_breadcrumbs.erb
JPrevost: Maybe. There are only two legit locations a user can be. The root url or the submission url. I suppose the normal behavior of a breadcrumb is to not display the link from the submission page in this case.

test/integration/authentication_test.rb
JPrevost: I feel pretty medium about explicitly testing a page that has no access restrictions to make sure it has no access restrictions.

Since I'll be making changes to this PR, I'll go ahead and add one though as I don't feel strongly against it either. :)

app/views/layouts/_breadcrumbs.erb
JPrevost: will fix

app/views/layouts/_breadcrumbs.erb
JPrevost: will fix

test/integration/authentication_test.rb
JPrevost: Added in static controller test

app/views/layouts/_breadcrumbs.erb
JPrevost: Updated to not show the link when on the new thesis page.

app/views/layouts/_breadcrumbs.erb
JPrevost: fixed

JPrevost:
@thatandromeda I've address your feedback (thx!). Let me know if this looks okay now and I'll squash it a bit.

app/views/layouts/_breadcrumbs.html.erb
```diff
@@ -3,7 +3,11 @@
     <div class="breadcrumbs">
       <%= link_to('Home', root_path) %>
       &raquo;
-      <%= link_to('Thesis Submission', new_thesis_path) %>
+        <% if current_page?(new_thesis_path) %>
+          Thesis Submission
+        <% else %>
+          <%= link_to('Thesis Submission', new_thesis_path) %>
```
thatandromeda: My understanding is that breadcrumbs are supposed to show where users are in the navigational hierarchy (so if people aren't on the thesis page, this shouldn't be a link to it, but rather an indication of where they are now).

Do we actually need breadcrumbs at all on a site with only like 3 pages, though, come to think of it? All of which are represented in the nav menu?

JPrevost: I'm not sure we actually have any other obvious navigation, but definitely agree this is a non-traditional (and backwards) use of breadcrumbs. I'll open a ticket to have UX weigh in on the general navigation and call this out specifically.

JPrevost: I'm going to remove the breadcrumbs and add pill nav.

thatandromeda: ++ on removing breadcrumbs (we don't need them in every app just because we have the css, right?)

We have the buttons in the local app header. Given that there are only 3 pages on the whole site I think they cover it.

JPrevost: There's no obvious way for a user on the home page to get to the submission page. I'll add that and then we're good.

JPrevost:
@thatandromeda latest commit removes breadcrumbs and adds pill navigation as suggested by fbot

app/views/layouts/_site_header.html.erb
```diff
@@ -23,6 +23,8 @@
     <div class="wrap-local-nav">
       <div class="wrap-bar">
         <nav class="local-nav" aria-label="Main menu">
+          <%= link_to("Home", root_path, class: 'button-secondary') %>
+          <%= link_to("Submit Thesis", new_thesis_path, class: 'button-secondary') %>
```
thatandromeda: 👍

PR #99, opened 2018-03-22
Initial hints and errors

JPrevost:
## Status
**READY**

#### What does this PR do?

Adds client side form validation and hints.

Look at the form.

- https://mitlibraries.atlassian.net/browse/TI-82

NO

YES -- sort of, pulling in a CDN version of jQuery Validate so no need to bundle but technically a new external dependency is being added

### DIFF
--- a/app/views/thesis/new.html.erb
+++ b/app/views/thesis/new.html.erb
@@ -3,46 +3,66 @@
   <div class="col3q">
     <h3 class="hd-2 title">Submit your thesis</h3>
     <%= simple_form_for(@thesis, url: '/thesis',
-             html: { class: 'thesisSubmission' } ) do |f| %>
+             html: { class: 'thesisSubmission' }, validate: true ) do |f| %>
 
     <%= f.error_notification %>
 
     <%= f.input :title, label: 'Title',
                         wrapper_html: { class: 'field-wrap' },
-                        input_html: { class: 'field field-text wide' } %>
+                        input_html: { class: 'field field-text wide',
+                          data: { msg: 'Your thesis title is required.'}
+                        } %>
 
     <%= f.input :abstract, label: 'Abstract', as: :text,
                            wrapper_html: { class: 'field-wrap' },
-                           input_html: { class: 'field field-text wide' } %>
+                           input_html: { class: 'field field-text wide',
+                             data: { msg: 'Your thesis abstract is required.'}
+                           } %>
 
-    <%= f.association :departments, label_method: :name, as: :check_boxes,
+    <%= f.association :departments, label_method: :name, as: :select,
+                         validate: { presence: true },
                          include_hidden: false,
-                         wrapper_html: { class: 'field-wrap list-checkboxes' },
-                         input_html: { class: 'field-checkbox' } %>
+                         wrapper_html: { class: 'field-wrap select' },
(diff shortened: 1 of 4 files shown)


app/views/thesis/new.html.erb
```diff
@@ -3,55 +3,74 @@
...
-                         wrapper_html: { class: 'field-wrap list-checkboxes' },
-                         input_html: { class: 'field-checkbox' } %>
+                         wrapper_html: { class: 'field-wrap select' },
+                         input_html: { class: 'field-select', multiple: false,
+                           data: { msg: 'Selecting your primary Department is required.'}
+                         },
+                         hint: 'Select your primary department as listed on your title page' %>
 
-    <%= f.association :degrees, label_method: :name, as: :check_boxes,
+    <%= f.association :degrees, label_method: :name, as: :select,
```
frrrances: 💯

app/views/thesis/new.html.erb
```diff
@@ -3,55 +3,74 @@
...
-                         input_html: { class: 'field-checkbox' } %>
-
-    <%= f.association :right, label_method: :statement, as: :radio_buttons,
-                              wrapper_html: { class: 'field-wrap list-radios' },
-                              input_html: { class: 'field-checkbox' } %>
+                         wrapper_html: { class: 'field-wrap select' },
+                         input_html: { class: 'field-select', multiple: false,
+                           data: { msg: 'Selecting your primary Degree is required.'}
+                         },
+                         hint: 'Select your primary degree as listed on your title page' %>
```
frrrances: Argh

app/views/thesis/new.html.erb
```diff
@@ -3,55 +3,74 @@
...
+                               collection: Thesis::VALID_MONTHS,
+                               wrapper_html: { class: 'field-wrap' },
+                               input_html: { class: 'field field-select', data: { msg: 'Selecting Month of graduation is required.' } } %>
 
-        <%= f.input :graduation_year, label: 'Year', as: :string,
-                            wrapper_html: { class: 'field-wrap' },
-                            input_html: { class: 'field field-select' } %>
+          <%= f.input :graduation_year, label: 'Year', as: :string,
+                              wrapper_html: { class: 'field-wrap' },
+                              input_html: { class: 'field field-select', data: { msg: 'Selecting Year of graduation is required.' } } %>
```
frrrances: "Entering" instead of "Selecting" - and can we have it validate that it's a number or is that more trouble than it's worth?

app/views/thesis/new.html.erb
```diff
@@ -3,55 +3,74 @@
...
       </div>
+      <span class="hint">For the degree date, enter the semester in which your degree will be conferred (typically your graduation semester).</span>
```
frrrances: Does this hint make sense with the values being months?

app/views/thesis/new.html.erb
JPrevost: It does on the backend but we should up front too. Will fix.

app/views/thesis/new.html.erb
JPrevost: I have no idea. I copy pasted from the old system lol

frrrances:
Sweet. Thanks for the info.  👍

PR #102, opened 2018-03-23
Updates pr build pattern match to allow fake auth

JPrevost:
## Status
**READY**

#### What does this PR do?

The PR builds now have incremented over 99 which seems to have
triggered a pattern match problem as they removed an extra character
from the middle of the expected URL to allow for the extra character
in the number to fit in their url size.

This should keep us safe for another 900 builds or so.

NO

NO


JPrevost:
@thatandromeda FYI: I'm not sure what heroku will do when we loop over 999 PR builds so I am not trying to solve the pattern match beyond our current problem. This will prob break in a few years. 🤷‍♀️

PR #104, opened 2018-03-26
Bugfix dept degree

JPrevost:
## Status
**READY**

#### What does this PR do?

There are two commits. The first adds a test for the _before_ behavior that was never tested in an Integration test.

The second, updates the behavior to reflect the change made in the UI which removed the multi select options for dept/degree. The controller was still expecting an array of values for those UI elements and thus was broken. The commit changes the controller expectation to be single values for each and updates tests in a couple locations to match that new behavior. The change to the UI that caused this bug landed last week. Ooops (on me).

NO

NO

### DIFF
--- a/app/controllers/thesis_controller.rb
+++ b/app/controllers/thesis_controller.rb
@@ -82,7 +82,7 @@ class ThesisController < ApplicationController
   def thesis_params
     params.require(:thesis).permit(:title, :abstract, :graduation_month,
                                    :graduation_year, :right_id,
-                                   department_ids: [], degree_ids: [])
+                                   :department_ids, :degree_ids)
   end
 
   def handle_thesis_ajax(handler)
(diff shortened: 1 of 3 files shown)


thatandromeda:
Wait, why did we remove the multiselect? I know that we do in fact have theses that belong to multiple departments (what with having pulled the entirety of dspace into hamlet, I had to design my database schema around this realization).

thatandromeda:
💯 on adding expected-behavior test and then breaking it with new behavior, though.

JPrevost:
The schema isn't changing. The UI was going to be awful once all the combinations were loaded so this is a concession between we don't actually _use_ the depts/degrees at this time but sort of want to later or something.

thatandromeda:
Can people still us the admin interface to set up multiple departments/degrees if worried users  @email them?

JPrevost:
Yup

PR #120, opened 2018-04-10
add local nav current state

frrrances:
## Status
**READY**

#### What does this PR do?
Updates style guide to add current state to local nav.


JPrevost:
@frrrances this doesn't fix the problem right, it just gives me the styles necessary to fix the problem?

frrrances:
@JPrevost exactly - once the HTML is updated, this will visually show the current page in the local nav.

PR #122, opened 2018-04-10
add screen reader and visual class for nav elements

JPrevost:
## Status
**READY/**

#### What does this PR do?

add screen reader span  and visual class for nav elements

Click through nav on pr app. Note you can tell which page you are on in the nav. Inspect elements to see the sr span.

- https://mitlibraries.atlassian.net/browse/TI-93

NO

NO

### DIFF
--- a/app/helpers/link_helper.rb
+++ b/app/helpers/link_helper.rb
@@ -1,0 +1,22 @@
+module LinkHelper
+  # helper wrapper around link_to that inserts visual and screen reader info
+  # for navigation links
+  def nav_link_to(name, url)
+    link_name = link_name(name, url)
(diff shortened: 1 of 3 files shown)


JPrevost:
I'm asking for a double review.

@frrrances can you confirm this does the a11y stuff we wanted.
@thatandromeda can you confirm this doesn't suck while doing the a11y stuff we wanted.

Thx :)

app/helpers/link_helper.rb
```diff
@@ -0,0 +1,22 @@
+module LinkHelper
```
thatandromeda: ++ for making it super easy for us to be accessible consistently.

PR #129, opened 2018-04-19
Orders departments alphabetically on ingest form

JPrevost:
## Status
**READY**

#### What does this PR do?

Updated display order of departments to be alphabetical. Stakeholders confirmed Degrees do not need the same ordering for some reason. (I suspect we'll need a way to more explicitly control ordering for that at some point as it appears random but they like it so it's clearly not).

In the PR build, add some additional departments in a non-alpha order. They should be alpha on the ingest form dropdown select.

- https://mitlibraries.atlassian.net/browse/TI-96

NO

NO

### DIFF
--- a/app/views/thesis/new.html.erb
+++ b/app/views/thesis/new.html.erb
@@ -25,6 +25,7 @@
                            } %>
 
     <%= f.association :departments, label_method: :name, as: :select,
+                         collection: Department.order(:name),
                          validate: { presence: true },
                          include_hidden: false,
                          wrapper_html: { class: 'field-wrap select' },


thatandromeda:
This works if our departments are named things like "Electrical Engineering & Computer Science" and "Aeronautics and Astronautics". It does not work if they are named "Course 6" and "Course 16", or "Course VI" and "Course IX" - both of these patterns will sort out of logical order.

Do we feel reasonably certain that the system will offer department name patterns like the first, and that thesis depositors are happy with that pattern rather than the latter two?

PR #138, opened 2018-04-23
Display accessible error banner on validation fail

JPrevost:
## Status
**READY**

#### What does this PR do?

Adds banner when client side form validation fails.

- https://mitlibraries.atlassian.net/browse/TI-94

NO

NO

### DIFF
--- a/app/views/thesis/new.html.erb
+++ b/app/views/thesis/new.html.erb
@@ -11,6 +11,7 @@
              html: { class: 'thesisSubmission' }, validate: true ) do |f| %>
 
     <%= f.error_notification %>
+    <div class="alert alert-banner error" style="display: none;" role="alert" aria-invalid="true"></div>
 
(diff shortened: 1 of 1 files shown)


app/views/thesis/new.html.erb
```diff
@@ -102,7 +103,22 @@
...
   });
   </script>
-  <script>$("#new_thesis").validate();</script>
+  <script>
+    $("#new_thesis").validate({
+      invalidHandler: function(event, validator) {
+        var errors = validator.numberOfInvalids();
+        if (errors) {
+          var message = errors == 1
+            ? 'You missed 1 field. It has been highlighted'
```
frrrances: The form was not submitted successfully - one required field needs to be fixed. 
The form was not submitted successfully - X required fields need to be fixed.

JPrevost: will change then merge. Thx!

PR #140, opened 2018-04-24
updated style guide with aligned inline form fields

frrrances:
## Status
**READY**

#### What does this PR do?
Makes inline form fields align better.

- https://mitlibraries.atlassian.net/browse/TI-107


JPrevost:
screen shot 2018-04-24 at 3 44 16 pm

PR #150, opened 2018-05-01
Removes activestorage config causing issues after 5.2 final

JPrevost:
## Status
**READY**

#### What does this PR do?

We had a fix in place for a RC build that was fixed in final and this removes that fix as our fix was actually breaking things now that it was fixed.

- https://mitlibraries.atlassian.net/browse/TI-108

NO

NO

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,7 +5,7 @@ Rails:
   Enabled: true
 
 AllCops:
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.4
   Exclude:
     - db/**/*
     - config/**/*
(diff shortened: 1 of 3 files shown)


.rubocop.yml
```diff
@@ -5,7 +5,7 @@ Rails:
   Enabled: true
 
 AllCops:
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.4
```
JPrevost: FYI. this is not a change in the ruby version but instead a change in the rubocop rules to reflect the ruby version we were already using.

thatandromeda: 👍 thanks for the FYI

PR #170, opened 2018-05-22
Sentry

JPrevost:
## Status
**READY**

#### What does this PR do?

Removes rollbar and adds sentry

NOTE: the PR build has an intentionally broken AWS S3 configuration so you can see the js error is caught, sent to sentry, but still shows users a fairly useful message in context.

NO

YES


app.json
```diff
@@ -39,11 +39,10 @@
     "RAILS_SERVE_STATIC_FILES": {
       "required": true
     },
-    "ROLLBAR_ACCESS_TOKEN": {
+    "SECRET_KEY_BASE": {
```
thatandromeda: ALL your secret key base are....oh never mind

app/assets/javascripts/direct_uploads.js
```diff
@@ -2,6 +2,8 @@
 /* original source:
```
thatandromeda: Is this comment still accurate?

JPrevost: yes, we've made a line or two of tweaks but it's largely copy/pasta from those docs. 👍

PR #404, opened 2020-01-31
Update dependencies

jazairi:
## Status
READY

#### What does this PR do?
Updates all app dependencies.

#### Todo:
- [x] Tests
- [ ] Documentation
- [ ] Stakeholder approval

YES

YES


jazairi:
@JPrevost I just updated the gems that depfu found individually, but `bundle update` found 11 more minor and patch updates. Should I push those as well? The tests pass.

JPrevost:
@jazairi Yeah it's totally cool to update as much as works. I'll show you my process for sanity checking this app in the PR build on Monday :)

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_02_27_190524) do
+ActiveRecord::Schema.define(version: 2018_02_27_185947) do
```
JPrevost: I'm curious if you know why this was needed? I'm a bit confused by it. I don't think it hurts anything but it also seems odd.

jazairi: Nope, I'm as confused as you are. Let's talk about it today.

jazairi: @JPrevost Am I clear to resolve this? My understanding is that we determined the timestamp is correct, despite its weirdness.

JPrevost: Yeah I think we're good. I'll resolve the convo. Thanks for following up to confirm.

PR #524, opened 2020-10-23
updates to use new heroku pr build naming pattern

JPrevost:
## Status
**READY**

#### What does this PR do?

updates or sanity check for fake auth to use new pr build naming parameters for the updated pipelines

NO

YES


matt-bernhardt (commented):
Is this a change that prevents failures (because the testing will look for invalid domains), or is it more a case of making sure that our naming conventions are accurate?

matt-bernhardt (approved):
Yep - that looks like the new naming convention. I see the tests passing, and the review app deployed as expected. :shipit:

PR #535, opened 2020-11-09
Update app for Rails 6

JPrevost:
Updates files changed from rails app:update to reflect both new settings
as well as to make sense with our existing application

Addresses deprecation warnings, but has not yet changed to new Rails 6
defaults.

Skips one test in which the data model had never checked for attached
files. It's unclear why this passed pre-Rails 6, but as upcoming ETD
features will require us to make Files optional for Thesis so we can
create stub records and then attach Files from Transfers, it's probably
best to not fix this just to have to unfix it later.

- https://mitlibraries.atlassian.net/browse/ENGX-48

YES

YES


app/models/thesis.rb
```diff
@@ -55,6 +55,8 @@ class Thesis < ApplicationRecord
   validates :degrees, presence:
     { message: VALIDATION_MSGS[:degrees] }
 
+  # validates :files, presence: true
```
JPrevost, line 58: Note. I put this here as an example of what you can now do with ActiveStorage in Rails 6 but that breaks a ton of tests that didn't consider this requirement of the model as they were built without it. I have confirmed even on the Rails 5.2 branch that if you just disable the JavaScript check in the uploader that ensures a file is attached it will submit it so this validation has never happened at the model level. We should remove this before merging but I wanted to make it easy for a reviewer to uncomment this locally to see what's going on if they wanted.

The new ETD features we are building will require this validation to be removed even if we did make it work properly now so it is legit not worth our time to fix this now.

jazairi (approved):
This all looks good. I agree with the decision to skip that test rather than delete it, since I assume we'll probably want to test that functionality as part of the upcoming ETD work.

PR #555, opened 2020-12-04
Document current production branch plan

JPrevost:

* The application is currently undergoing major feature development
* We need a way to maintain the production application without making
  work on the new features more complicated

* This documents the new `1.x` branch that has been created with notes
  on how to update production during this feature development period.

Someone will need to remember to delete this note when it no longer
makes sense.

NO

NO

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,18 @@
 This is a simple web app to collect metadata and files from a `User` and allow
 `Users` with the role `Admin` to download and edit the metadata.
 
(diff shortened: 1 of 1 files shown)


jazairi (approved):
:shipit:, but we should probably make a note somewhere (JIRA ticket?) to update the readme when the project concludes. I can easily imagine myself forgetting to remove this bit. In fact, I already have.

PR #556, opened 2020-12-04
Adds Transfer model and relevant tests

matt-bernhardt:
This builds out the Transfer model as provided by Helen in ETS-62. It includes tests cribbed from the Thesis model, with the addition of some tests specific to attached files.

This is the first of several tickets that will define the bulk transfer capability.

The easiest way to see this change will be via the tests, or using a local console to create/update Transfers. The UI is unaffected.

https://mitlibraries.atlassian.net/browse/ETS-41

#### Todo:
- [x] Tests

YES

NO

### DIFF
--- a/app/models/transfer.rb
+++ b/app/models/transfer.rb
@@ -1,0 +1,69 @@
+# == Schema Information
+#
+# Table name: transfers
+#
+#  id            :integer          not null, primary key
+#  user_id       :integer          not null
+#  department_id :integer          not null
+#  grad_date     :date             not null
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#
+class Transfer < ApplicationRecord
+  belongs_to :user
+  belongs_to :department
+
+  has_many_attached :files
+
+  attr_accessor :graduation_year, :graduation_month
+
+  VALIDATION_MSGS = {
+    graduation_year: 'Required - Please input your year of graduation.',
+    graduation_month: 'Required - Please select your month of graduation.',
+    files: 'Required - Attaching at least one file is required.',
+  }
+
+  validates :graduation_year, presence:
+    { message: VALIDATION_MSGS[:graduation_year] }
+  validate :valid_year?
+  validates :graduation_month, presence:
+    { message: VALIDATION_MSGS[:graduation_month] }
+  validate :valid_month?
+  validates :files, presence: true
+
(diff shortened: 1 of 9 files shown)


test/models/transfer_test.rb
```diff
@@ -0,0 +1,136 @@
...
+
+class TransferTest < ActiveSupport::TestCase
+  setup do
+    @transfer = transfers(:valid)
+    file = Rails.root.join('test','fixtures','files','a_pdf.pdf')
+    @transfer.files.attach(io: File.open(file), filename: 'a_pdf.pdf')
+  end
+
+  teardown do
+    Rails.cache.clear
```
JPrevost, line 22: Is the intent of this to delete the ActiveStorage file attached in `setup`? If so, I'm not sure clearing the general Rails cache will achieve that.

test/fixtures/transfers.yml
```diff
@@ -0,0 +1,20 @@
...
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#
+valid:
+  department: one
+  grad_date: 2020-05-01
+  user: yo
+
+baduser:
+  department_id: 0
```
JPrevost, line 18: I believe this fixture should use the format of `department: name_of_department_fixture` instead of a specific `department_id`.

matt-bernhardt: Ah, yeah - good catch. I'll fix.

test/models/transfer_test.rb
matt-bernhardt: Ok, let me poke on what a better way to do this would be...

app/models/transfer.rb
```diff
@@ -0,0 +1,69 @@
...
+#  id            :integer          not null, primary key
+#  user_id       :integer          not null
+#  department_id :integer          not null
+#  grad_date     :date             not null
+#  created_at    :datetime         not null
+#  updated_at    :datetime         not null
+#
+class Transfer < ApplicationRecord
+  belongs_to :user
+  belongs_to :department
```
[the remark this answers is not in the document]
matt-bernhardt: Okay, I've added the relationship to the two models, as well as tests that we can use the relationship to access transfers from those models.

JPrevost (approved):

PR #558, opened 2020-12-10
ETS-71 Add submitter model

jazairi:

We need a Submitter model to know which Users can submit Transfers
for various Departments.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETS-71

Adds Submitter model to join Transfers and Departments.

New database migration.

NO

### DIFF
--- a/app/models/department.rb
+++ b/app/models/department.rb
@@ -12,6 +12,8 @@ class Department < ApplicationRecord
   has_many :department_theses
   has_many :theses, through: :department_theses
   has_many :transfers
+  has_many :submitters
+  has_many :users, through: :submitters
 
   validates :name, presence: true
 end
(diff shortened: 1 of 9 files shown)


matt-bernhardt (commented):
The data model changes are in accord with the diagrams that Helen drafted, and the tests are sensible.

My question - which is probably more about the ticket than this work - is why we're naming this table `submitter` rather than a more explictly-linking name like `department_user`? This would match the existing `department_theses` table, and make it clearer (IMO) from an outside perspective that the table holds no internal data, that it only provides links between the other two tables.

matt-bernhardt (approved):
Having had the discussion on Slack about the justification for naming this model `Submitter`, I willing to :shipit: this. I understand that this is a table in support of user rights, but that using the `role` field in the user table isn't adequate because individual users will have rights for only certain departments.

Thanks for hearing out my concerns on this - let's merge this and move on :-)

matt-bernhardt (approved):
Once more, this time actually approving...

PR #565, opened 2020-12-15
ETS-40 Add transfer submitter role

jazairi:

The Ability model needs a submitter role to control who has
permission to deposit and manage transfers.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETS-40

This adds the transfer submitter role, along with the preliminary
TransferController with create and show methods. (Additional logic
will be added to the controller as needed.)

The Ability model now includes the following additional permissions:
* thesis_admin can create transfers and view any transfer
* transfer_submitter can create transfers and view their own transfers
* processor can view any transfer

* Generated the transfer dashboard and created a couple of stubbed
transfer views.
* Minor changes to our test fixtures to make the controller tests
clearer.
* Updated a thesis controller test assertion to redirect to the
login URL specifically, instead of redirecting anywhere.

#### Todo:
- [x] Tests
- [x] Documentation
- [x] Stakeholder approval

NO

NO

### DIFF
--- a/test/controllers/thesis_controller_test.rb
+++ b/test/controllers/thesis_controller_test.rb
@@ -4,7 +4,7 @@ class ThesisControllerTest < ActionDispatch::IntegrationTest
   # ~~~~~~~~~~~~~~~~~~~~~~~~~~ the submission system ~~~~~~~~~~~~~~~~~~~~~~~~~~
   test 'new prompts for login' do
     get '/thesis/new'
-    assert_response :redirect
+    assert_redirected_to '/users/auth/saml'
   end
 
   test 'new when logged in' do
--- a/app/controllers/admin/transfers_controller.rb
+++ b/app/controllers/admin/transfers_controller.rb
@@ -1,0 +1,46 @@
+module Admin
+  class TransfersController < Admin::ApplicationController
+    # Overwrite any of the RESTful controller actions to implement custom behavior
+    # For example, you may want to send an email after a foo is updated.
+    #
+    # def update
+    #   super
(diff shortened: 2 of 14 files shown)


jazairi:
Looks like the decrease in coverage is due to the new transfer dashboard. My assumption is that this wouldn't require coverage since it doesn't do anything yet, but I could be mistaken about that.

test/controllers/thesis_controller_test.rb
```diff
@@ -4,7 +4,7 @@ class ThesisControllerTest < ActionDispatch::IntegrationTest
   # ~~~~~~~~~~~~~~~~~~~~~~~~~~ the submission system ~~~~~~~~~~~~~~~~~~~~~~~~~~
   test 'new prompts for login' do
     get '/thesis/new'
-    assert_response :redirect
+    assert_redirected_to '/users/auth/saml'
```
matt-bernhardt, line 7: Is this change related to anything else in the PR? I see that this matches what the test is for the transfer controller - so I'm reading this as "what the test should have been" rather than a substantive change. I'm fine with that, but don't want to miss anything.

jazairi: Yes, this change is unrelated to the goal of the PR. I noticed that the test doesn't really do what it seeks to, so I figured I'd fix it as I was writing similar tests for the TransferController.

matt-bernhardt (approved):
This all seems to be fine to me. The tests pass, and there's no obvious edge cases missed. Some of this will get its UI started in the next ticket, while other aspects may come a bit further down the road.

:shipit:

PR #572, opened 2020-12-18
ETS-33 Adds a basic UI for Transfers, including integration tests

matt-bernhardt:
** Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETS-33

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/views/transfer/new.html.erb
+++ b/app/views/transfer/new.html.erb
@@ -10,7 +10,7 @@
   <div class="alert alert-banner error" style="display: none;" role="alert" aria-invalid="true"></div>
 
   <%= f.association :department, label_method: :name, as: :select,
-        collection: current_user.departments.order(:name),
+        collection: current_user.submittable_departments,
         validate: { presence: true },
         include_hidden: false,
         wrapper_html: { class: 'field-wrap select' },
--- a/app/views/transfer/new.html.erb
+++ b/app/views/transfer/new.html.erb
@@ -1,1 +1,100 @@
 <%= content_for(:title, "Thesis Transfer Submission | MIT Libraries") %>
+
+<% content_for :additional_js do %>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery-validate/1.17.0/jquery.validate.min.js"></script>
+<% end %>
+
+<%= simple_form_for(@transfer, url: '/transfer', html: { class: 'transferSubmission' }, validate: true ) do |f| %>
+
+  <%= f.error_notification %>
+  <div class="alert alert-banner error" style="display: none;" role="alert" aria-invalid="true"></div>
+
+  <%= f.association :department, label_method: :name, as: :select,
+        collection: current_user.departments.order(:name),
+        validate: { presence: true },
+        include_hidden: false,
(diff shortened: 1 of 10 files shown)


app/views/transfer/new.html.erb
```diff
@@ -1 +1,100 @@
 <%= content_for(:title, "Thesis Transfer Submission | MIT Libraries") %>
+
+<% content_for :additional_js do %>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery-validate/1.17.0/jquery.validate.min.js"></script>
```
JPrevost, line 4: 🙈 (and yes, I know this is how we do this already in this app so it is totally appropriate but also 🙈 )

matt-bernhardt: Heh - I've been faced with this exact problem in some personal projects, and noted that this is how it happens here. From a design perspective, it would be interesting to discuss a better way of doing this at some point.

But yes, a bit of 🙈 for now.

app/views/transfer/new.html.erb
```diff
@@ -1 +1,100 @@
...
+<% end %>
+
+<%= simple_form_for(@transfer, url: '/transfer', html: { class: 'transferSubmission' }, validate: true ) do |f| %>
+
+  <%= f.error_notification %>
+  <div class="alert alert-banner error" style="display: none;" role="alert" aria-invalid="true"></div>
+
+  <%= f.association :department, label_method: :name, as: :select,
+        collection: current_user.departments.order(:name),
```
JPrevost, line 13: I think this should be the default for this field, but for `theses_admins` I believe it should just display _all_ Departments. If that is a pain to make that change (and associated tests) I would be okay with opening a specific ticket to address that distinction. The current solution will technically not restrict admins from submitting, but the form will never allow them to choose anything for this field so effectively they still can't submit :)

matt-bernhardt: Hmm... ok. I had thought that, for users with the `:thesis_admin` ability, this call would end up returning all departments because their ability would be for all departments. If we want to structure this differently, that's fine - I can make this change and its test.

JPrevost: The way it's modeled we need to handle that in the UI here or would need more complex behavior in the models. I normally move towards complexity in the models so maybe we should talk about your expectation as it may be a very good one.

JPrevost (changes requested):
If we can roll in the change to allow thesis_admins to submit to any dept I think that would be great to include, but if it gets messy we can open a new ticket.

I'd like to hold off on merging until we get stakeholder feedback if we can.

matt-bernhardt:
I'm working on the change to how thesis_admins see the list of available departments, but this will likely to take until Monday at this point - what should be a very short change has turned into 5-7 failing tests. I'm blaming Friday PM brain.

I agree with getting stakeholder approval, so I think this buys me time to work out this last bit.

app/views/transfer/new.html.erb
matt-bernhardt: I think this is resolved now...

JPrevost (approved):
🌈

PR #583, opened 2021-01-13
Adds models for holds and hold_sources

matt-bernhardt:
This adds models for Holds and Hold_Sources, along with basic tests.

The details of these models are derived from the 1.2 data model that can be found linked to the Jira ticket in this PR.

#### Developer

#### Code Reviewer

YES

NO

### DIFF
--- a/app/models/hold.rb
+++ b/app/models/hold.rb
@@ -15,8 +15,13 @@
 #  updated_at       :datetime         not null
 #
 class Hold < ApplicationRecord
-	belongs_to :thesis
-	belongs_to :hold_source
+  belongs_to :thesis
+  belongs_to :hold_source
 
-	enum status: [ :active, :expired, :released ]
+  enum status: [ :active, :expired, :released ]
+
+  validates :date_requested, presence: true
+  validates :date_start, presence: true
+  validates :date_end, presence: true
+  validates :status, presence: true
 end
--- a/app/models/hold.rb
+++ b/app/models/hold.rb
@@ -1,0 +1,22 @@
+# == Schema Information
+#
+# Table name: holds
+#
+#  id               :integer          not null, primary key
+#  thesis_id        :integer          not null
+#  date_requested   :date             not null
+#  date_start       :date             not null
+#  date_end         :date             not null
+#  hold_source_id   :integer          not null
+#  case_number      :string
+#  status           :integer          not null
+#  processing_notes :text
+#  created_at       :datetime         not null
+#  updated_at       :datetime         not null
+#
+class Hold < ApplicationRecord
+	belongs_to :thesis
+	belongs_to :hold_source
+
+	enum status: [ :active, :expired, :released ]
+end
--- a/app/models/hold_source.rb
+++ b/app/models/hold_source.rb
@@ -8,5 +8,7 @@
 #  updated_at :datetime         not null
 #
 class HoldSource < ApplicationRecord
-	has_many :holds
+  has_many :holds
+
+  validates :source, presence: true
 end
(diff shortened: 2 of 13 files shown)


app/models/hold.rb
```diff
@@ -0,0 +1,22 @@
...
+#  case_number      :string
+#  status           :integer          not null
+#  processing_notes :text
+#  created_at       :datetime         not null
+#  updated_at       :datetime         not null
+#
+class Hold < ApplicationRecord
+	belongs_to :thesis
+	belongs_to :hold_source
+
```
JPrevost, line 20: Adding validations for each required element in the models would be better here and then associated tests to confirm our business rules. The seemingly large amount of tests in this case won't be testing Rails so much as making sure we notice when someone changes the model validations in a way that should break our understanding of the business rules as defined in our tests, we should find out by tests failing.

Right now, I can create a record with no date_end, no date_start, no  date_requested and when I check `.valid?` says true but when I save it it fails with a db constraint error.

```
#<Hold:0x00007fef07641e70
 id: nil,
 thesis_id: 43,
 date_requested: nil,
 date_start: nil,
 date_end: nil,
 hold_source_id: 1,
 case_number: "123456",
 status: "active",
 processing_notes: "Hallo!",
 created_at: Wed, 13 Jan 2021 19:26:44 UTC +00:00,
 updated_at: Wed, 13 Jan 2021 19:26:44 UTC +00:00>
```

☝🏻 that's clearly an invalid Hold by our standards but rails thinks it is valid and will try to save it (and then fail).

app/models/hold_source.rb
```diff
@@ -0,0 +1,12 @@
+# == Schema Information
+#
+# Table name: hold_sources
+#
+#  id         :integer          not null, primary key
+#  source     :text             not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
```
JPrevost, line 9: Similarly to the Holds mode, please add validations to this model to allow Rails to know when an Object is valid and not. You can see why quickly in this model in the `rails console` by:
hs = HoldSource.new
hs.valid?

That will be true. It is not :)

test/models/hold_test.rb
```diff
@@ -0,0 +1,41 @@
...
+#  updated_at       :datetime         not null
+#
+require 'test_helper'
+
+class HoldTest < ActiveSupport::TestCase
+  test 'valid hold' do
+    hold = holds(:valid)
+    assert(hold.valid?)
+  end
+
```
JPrevost, line 24: See Holds model comment about adding tests for all of the not nullable fields.

test/models/hold_source_test.rb
```diff
@@ -0,0 +1,17 @@
...
+#
+# Table name: hold_sources
+#
+#  id         :integer          not null, primary key
+#  source     :text             not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+require 'test_helper'
+
```
JPrevost, line 11: See HoldSource for comment about adding tests for not nullable fields.

JPrevost (commented):
See comments about some required changes around model validations and a couple small things that can happen later under the UI ticket or spin into new tickets.

app/models/hold_source.rb
```diff
@@ -0,0 +1,12 @@
...
+class HoldSource < ApplicationRecord
```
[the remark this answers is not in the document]
matt-bernhardt: Adding the values from the data model to `db/seeds.rb` sounds like a perfect thing for this PR to tackle. I'm still debating whether the admin UI bits should be in this PR or the other UI one...stay tuned there but yes - if it doesn't land here it should be in the other one.

matt-bernhardt:
Okay @JPrevost - I think I've addressed the feedback you provided. Can you take another look now?

JPrevost (approved):
👍🏻

PR #592, opened 2021-01-28
Adds Primary field to Department-Thesis relationship

matt-bernhardt:
This adds the Primary flag to the Department_Thesis table. There are some associated changes with this ticket which are described in the commit message. I'd rather not type them all out here again, but will copy-paste if reviewers want.

Ticket: https://mitlibraries.atlassian.net/browse/ETD-64

#### Developer

#### Code Reviewer

YES

NO

### DIFF
--- a/test/integration/admin_dashboard_test.rb
+++ b/test/integration/admin_dashboard_test.rb
@@ -413,18 +413,17 @@ class AuthenticationTest < ActionDispatch::IntegrationTest
   end
 
   test 'can edit department_thesis through admin dashboard' do
-    needle = false
     mock_auth(users(:thesis_admin))
     link = DepartmentThesis.first
-    assert_not_equal needle, link.primary
+    assert_not_equal false, link.primary
     patch admin_department_thesis_path(link),
-      params: { department_thesis: { primary: needle } }
+      params: { department_thesis: { primary: false } }
     link.reload
-    assert_equal needle, link.primary
+    assert_equal false, link.primary
     patch admin_department_thesis_path(link),
-      params: { department_thesis: { primary: !needle } }
+      params: { department_thesis: { primary: true } }
     link.reload
-    assert_not_equal needle, link.primary
+    assert_not_equal false, link.primary
   end
 
   # Holds
--- a/test/integration/admin_dashboard_test.rb
+++ b/test/integration/admin_dashboard_test.rb
@@ -388,6 +388,45 @@ class AuthenticationTest < ActionDispatch::IntegrationTest
     assert_equal needle.title, advisor.theses.first.title
   end
 
+  # Department_Theses
+  test 'accessing department_thesis panel does not work with basic rights' do
+    mock_auth(users(:basic))
+    get '/admin/department_theses'
+    assert_response :redirect
+    mock_auth(users(:processor))
+    get '/admin/department_theses'
+    assert_response :redirect
+  end
+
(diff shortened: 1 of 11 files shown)


test/integration/admin_dashboard_test.rb
```diff
@@ -388,6 +388,45 @@ def teardown
...
+
+  test 'accessing department_thesis panel as a thesis_admin user works' do
+    mock_auth(users(:thesis_admin))
+    get '/admin/department_theses'
+    assert_response :success
+    assert_equal('/admin/department_theses', path)
+  end
+
+  test 'can edit department_thesis through admin dashboard' do
+    needle = false
```
JPrevost, line 416: This definitely works, so can stay. However, I wanted to share the needle confused me when reading this and I'm not yet convinced (but am happy to be swayed) that it's clearer with the needle variable than just using `true` and `false` directly in the code. Looking at `!needle` and then having to look through to follow the state of `needle` to ensure it didn't change kinda broke my Friday afternoon brain lol

test/models/department_thesis_test.rb
```diff
@@ -0,0 +1,42 @@
...
+  test 'valid link with and without primary' do
+    link = department_theses(:primary)
+    assert_equal link.primary, true
+    assert(link.valid?)
+    link = department_theses(:other)
+    assert_equal link.primary, false
+    assert(link.valid?)
+  end
+
+  test 'only one primary department per thesis' do
```
JPrevost, line 23: Looking at this test doesn't make me confident without looking at the underlying data that the test does what it says. Would you consider making it clearer in the test what is going on?

JPrevost (changes requested):
My code comments are non blocking.

However, when testing this in the PR build if I look at Department Theses in administrate I get an error if I have submitted a Thesis via the Student submission form.

This may be expected due to that form not collecting data in a way that works with this data model change? I'm not sure so thought we should look into that before proceeding.

test/integration/admin_dashboard_test.rb
matt-bernhardt: Yeah, I can hear that feedback. I struggled a bit with whether I should use a variable for this. We use `needle` as a variable in tests several times throughout this file, but in this case it felt a bit awkward. Seeing your feedback on this makes me think that we should try it another way. Stay tuned...

test/models/department_thesis_test.rb
matt-bernhardt: Sure - looking at this again with fresh coffee, I can see that this is a bit sparse. I've got a longer test now that does more to explicitly test the underlying conditions prior to these actions.

test/integration/admin_dashboard_test.rb
JPrevost: Yeah. To be clear I'm not opposed to a variable (named `needle` or not) in general. Just for booleans it was making me think harder than I wanted to lol

matt-bernhardt:
The problems in the review app were apparently brought about by my squashing some earlier migrations down to one for the purposes of easier code review. Deleting the review app and re-creating it entirely caused the `id` field in the department_thesis table to be created as expected.

I've updated the tests to address the two comments that you made on Friday - that was good feedback, thanks.

JPrevost (approved):

matt-bernhardt:
Helen has approved this change against the data model - going to merge now...

PR #593, opened 2021-01-28
Adds fields to active_storage_attachment

JPrevost:

* Our workflows require the ability to track the reason a file is being
  submitted (primary thesis file, supplementary files, signature page,
  etc)

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-30
* https://mitlibraries.atlassian.net/browse/ETD-145

* By extending the active_storage_attachment class and adding two fields
  to the associated database table, we can easily provide the info our
  app needs
* The extra module we load into the class is only necessary to allow an
  enum, but it has a nice side effect of allowing us to include the
  module with our models to be more visible to developers that there is
  something non-default in our active_storage_attachment class.

By extending the active_storage_attachment class, it is possible future
changes to ActiveStorage may be in conflict with our changes. This feels
quite unlikely, but as it's possible it feels worth noting.

We are using the terminology "purpose" for the "type" of file. The
reason being that a file "type" is commonly used to describe things like
"it's a PDF" or "it's a zip file". As we _also_ care about that meaning
of "type" in this application and already store that by default in
ActiveStorage it felt useful longer term to choose another label to
represent the "purpose" of the file being submitted.

#### How to see this change

The current UI does not have a way to set these new fields. It does have a rudimentary way to _view_ these fields in both Transfer and Thesis views. This is not intended as the final UI, but just a way to show the data _someplace_.

Locally, or on the PR build, you can manually populate these fields in the `rails console` doing something like:

```
t = Transfer.last
f = t.files.first
f.description = "hallo!"
f.purpose = 0
f.save

pp f (or view in the Transfer administrate panel)

#<ActiveStorage::Attachment:0x00007ff175f3e300
 id: 83,
 name: "files",
 record_type: "Transfer",
 record_id: 6,
 blob_id: 102,
 created_at: Tue, 26 Jan 2021 14:51:39 UTC +00:00,
 description: "hallo!",
 purpose: "thesis_pdf">
```

#### Developer

#### Code Reviewer

YES

NO

### DIFF
--- a/app/dashboards/thesis_dashboard.rb
+++ b/app/dashboards/thesis_dashboard.rb
@@ -26,7 +26,7 @@ class ThesisDashboard < Administrate::BaseDashboard
     graduation_year: Field::Number,
     created_at: Field::DateTime,
     updated_at: Field::DateTime,
-    files: FileField,
+    files: AttachmentField,
     files_complete: Field::Boolean,
     status: Field::Select.with_options(
       collection: Thesis::STATUS_OPTIONS,
--- a/app/fields/file_field.rb
+++ b/app/fields/file_field.rb
@@ -1,7 +1,0 @@
-require 'administrate/field/base'
-
-class FileField < Administrate::Field::Base
-  def files
-    data.record.files
-  end
-end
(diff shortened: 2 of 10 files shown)


jazairi (approved):
The changes work as expected in the rails console for the PR build. :shipit: 

One non-blocking question, possibly for a future ticket: Coveralls is mad about FileField, but we don't need this anymore, right? Should we remove it from the codebase?

JPrevost:
@jazairi good catch. I also forgot to document that side effect of this work as it was not required but when I was poking around I noticed we had two things that did the same thing and combined them. I'll remove the extra file and amend the commit message after I squash to document that was part of this change set.

PR #601, opened 2021-02-09
Updates naming convention used around fake auth

matt-bernhardt:
## Why are these changes being introduced:

* For our review applications we enable some faked authentication for
  easier testing. To prevent enabling this inadvertently on other tiers,
  we restrict its operation to a specific naming pattern. This pattern
  will change when we relaunch the application in a new domain name,
  so this pattern must be updated.

## Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-148

## How does this address that need:

* Updates the naming convention in the fake_auth concern
* Removes an unneeded test now that PR builds only follow one naming
  convention
* Updates the name of the staging and production apps, which prevents
  fake_auth from being enabled for those applications.

## Document any side effects to this change:

* This change *replaces* the naming pattern approved for review apps, so merging this branch will prevent `main` from being able to deploy to the old pipeline. I believe that this is acceptable, as the `1.x` branch is where we're maintaining the legacy application until it is retired.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/test/models/fake_auth_test.rb
+++ b/test/models/fake_auth_test.rb
@@ -6,6 +6,12 @@ class FakeAuthTest < ActiveSupport::TestCase
   test 'fakeauth disabled' do
     ClimateControl.modify(
       FAKE_AUTH_ENABLED: 'false',
+      HEROKU_APP_NAME: 'thesis-submit-pr-123'
+    ) do
+      assert_equal(false, fake_auth_status)
+    end
+    ClimateControl.modify(
+      FAKE_AUTH_ENABLED: 'false',
       HEROKU_APP_NAME: 'thesis-dropbox-pr-123'
     ) do
       assert_equal(false, fake_auth_status)
@@ -15,13 +21,10 @@ class FakeAuthTest < ActiveSupport::TestCase
   test 'fakeauth enabled pr pattern app name' do
     ClimateControl.modify(
       FAKE_AUTH_ENABLED: 'true',
-      HEROKU_APP_NAME: 'thesis-dropbox-pr-123'
+      HEROKU_APP_NAME: 'thesis-submit-pr-123'
     ) do
       assert_equal(true, fake_auth_status)
     end
-  end
-
-  test 'fakeauth enabled pr minus s pattern app name' do
     ClimateControl.modify(
       FAKE_AUTH_ENABLED: 'true',
       HEROKU_APP_NAME: 'thesis-dropbox-pr-123'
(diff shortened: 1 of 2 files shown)


test/models/fake_auth_test.rb
```diff
@@ -15,16 +15,7 @@ class FakeAuthTest < ActiveSupport::TestCase
   test 'fakeauth enabled pr pattern app name' do
     ClimateControl.modify(
       FAKE_AUTH_ENABLED: 'true',
-      HEROKU_APP_NAME: 'thesis-dropbox-pr-123'
-    ) do
-      assert_equal(true, fake_auth_status)
-    end
-  end
-
-  test 'fakeauth enabled pr minus s pattern app name' do
```
matt-bernhardt, line 24: Please check me on this - I believe that now, with only the `thesis-submit-pr-#` naming convention, we only need one test here. From the name, and the history of this test, it seems like we can delete this because `...pr-s-` is no longer a pattern we might encounter.

JPrevost: yeah we can control PR name patterns more now than we could then so I do suspect this is no longer relevant.

test/models/fake_auth_test.rb
```diff
@@ -6,7 +6,7 @@ class FakeAuthTest < ActiveSupport::TestCase
   test 'fakeauth disabled' do
     ClimateControl.modify(
       FAKE_AUTH_ENABLED: 'false',
-      HEROKU_APP_NAME: 'thesis-dropbox-pr-123'
+      HEROKU_APP_NAME: 'thesis-submit-pr-123'
```
JPrevost, line 9: Would you mind adding the old tests back as well as these updated ones? I know we'll rip some of this out when we retire the old pipeline, for now showing that both the old and new patters are working explicitly in tests with that regex would be great.

Also, I now wish we controlled fake auth patterns and pr build patterns in ENV but let us not go down that path now even if it would feel good...

matt-bernhardt: I think I've now done this - although for some reason the diff on this screen isn't looking right to me. The source code itself at https://github.com/MITLibraries/thing/blob/etd-148-fake-auth/test/models/fake_auth_test.rb#L6-L34 looks correct, with tests for both old and new review app names, but the combined diff looks like it isn't displaying all the relevant lines.

test/models/fake_auth_test.rb
matt-bernhardt: I've not restored this test yet, as I think it is still not needed even with the old and new pipelines being tested. If you'd like this back too, I can add it.

JPrevost: No you are correct, this still remains unneeded.

JPrevost (approved):

PR #611, opened 2021-02-17
ETD-175 Update date format in Hold and Hold Source UIs

jazairi:

The current administrate field (DateTime) for created_at and updated_at
includes the timestamp, which adds visual clutter. This change was
requested as part of the UX review of the Hold UI.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-175
* https://mitlibraries.atlassian.net/browse/ETD-106
* https://mitlibraries.atlassian.net/browse/ETD-170

This removes the timestamp from the DateTime fields in the Hold and
Hold Source UIs. The changes can be verified by confirming no timestamps
in the created_at and updated_at fields on the Hold show view and the Hold 
Source index and show views.

None.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/dashboards/hold_dashboard.rb
+++ b/app/dashboards/hold_dashboard.rb
@@ -17,8 +17,8 @@ class HoldDashboard < Administrate::BaseDashboard
     case_number: Field::String,
     status: Field::Select.with_options(searchable: false, collection: ->(field) { field.resource.class.send(field.attribute.to_s.pluralize).keys }),
     processing_notes: Field::Text,
-    created_at: Field::DateTime,
-    updated_at: Field::DateTime,
+    created_at: Field::Date,
+    updated_at: Field::Date,
   }.freeze
 
   # COLLECTION_ATTRIBUTES
--- a/app/dashboards/hold_source_dashboard.rb
+++ b/app/dashboards/hold_source_dashboard.rb
@@ -12,8 +12,8 @@ class HoldSourceDashboard < Administrate::BaseDashboard
     holds: Field::HasMany,
     id: Field::Number,
     source: Field::Text,
-    created_at: Field::DateTime,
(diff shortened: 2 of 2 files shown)


app/dashboards/hold_dashboard.rb
```diff
@@ -17,8 +17,8 @@ class HoldDashboard < Administrate::BaseDashboard
     case_number: Field::String,
     status: Field::Select.with_options(searchable: false, collection: ->(field) { field.resource.class.send(field.attribute.to_s.pluralize).keys }),
     processing_notes: Field::Text,
-    created_at: Field::DateTime,
-    updated_at: Field::DateTime,
+    created_at: Field::DateTime.with_options(format: "%a, %d %b %Y"),
```
JPrevost, line 20: I'm curious if we should match the display we have for the `DATE REQUESTED` etc which is `2021-02-18` instead of `Thu, 18 Feb 2021` for consistency within the same UI. Totally not blocking, but I sort of predict we'll eventually get that feedback so maybe we should just do it? I'm fine with merging as-is or normalizing in the same UI as there was no explicit stakeholder guidance I think

JPrevost (approved):
See comment about possibly tweaking display format but feel free to merge as-is if you disagree (or even if you agree but prefer to not make that change now).

app/dashboards/hold_dashboard.rb
jazairi: Yeah, that makes sense. I asked Stephanie about it yesterday, and since neither of us had a preference, I figured I would just remove the timestamp and leave any specific formatting requests up to Mikki. However, it's cleaner and easier to just make it `Field::Date`, so I think we can roll with that approach for now.

PR #614, opened 2021-02-18
Adds support for data uploads from Registrar

matt-bernhardt:
This creates a Registrar model to support uploading a CSV file that we'll get from the Registrar's office. There is an upload form, a (blank) show route, and test coverage to keep our current coverage level in place. I've confirmed that the uploaded file ends up in our s3 bucket.

Some notes on some specific choices I've made, in case any adjustments are needed:

* Both `admin` and `thesis_admin` roles have access to this form. I think I've written tests sufficient to clarify this, but I confess that at times the role system and admin flag trip me up.
* Jeremy commented on this already, but there is a user attribute on this model. I'm not sure if that's what's needed, but this may be a question for Helen.
* Do we need a MIME type check on the uploaded CSV file? The ticket mentions not doing anything with the uploaded file, and I'm not sure if this is part of that or not. I think I know how we'd do it if it's needed now, but didn't want to borrow trouble.

#### Developer

#### Code Reviewer

YES

NO

### DIFF
--- a/app/models/registrar.rb
+++ b/app/models/registrar.rb
@@ -1,0 +1,11 @@
+class Registrar < ApplicationRecord
+  belongs_to :user
+
+  has_one_attached :graduation_list
+
+  VALIDATION_MSGS = {
+    graduation_list: 'Required - Attaching a CSV file is required.',
+  }
+
+  validates :graduation_list, presence: true
+end
--- a/app/controllers/registrar_controller.rb
+++ b/app/controllers/registrar_controller.rb
@@ -1,0 +1,45 @@
+class RegistrarController < ApplicationController
+  before_action :require_user
+  before_action :authenticate_user!
+  load_and_authorize_resource
+  protect_from_forgery with: :exception
+
+  def new
+    @registrar = Registrar.new
+    @registrar.user = current_user
+  end
+
+  def create
+    @registrar = Registrar.new(registrar_params)
+    @registrar.user = current_user
+    @registrar.graduation_list.attach(params[:registrar][:graduation_list])
+    if @registrar.save
+      flash.notice = 'Thank you for submitting this Registrar file.'
+      redirect_to root_path()
+    else
+      flash[:error] = "Error saving Registrar file: #{@registrar.errors.full_messages}"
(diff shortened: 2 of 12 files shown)


app/models/registrar.rb
```diff
@@ -0,0 +1,11 @@
+class Registrar < ApplicationRecord
+  belongs_to :user
```
JPrevost, line 2: I know this isn't open for review officially, but I didn't notice that and looked so here's my feedback unofficially :)

- I don't think we want a user associated with these files unless someone else has provide use cases that contradict this. Were you thinking of a use case I may not be? I'm assuming we'll just process these and not care who uploaded them because only admins can upload them. I guess maybe keeping them around after processing for a period of time and knowing who did stuff isn't awful though so... maybe your way is actually better.
- We will need permissions added in ability.rb to ensure only thesis_admins can upload files.

I do think we need to have permissions added to ability

matt-bernhardt: Yeah, I've gone back and forth with myself about whether we need the user information or not. I tried doing without it initially, but ran into some issues that seemed to be forcing me into having this. I may not have looked closely enough, though. The current requirement doesn't list a user field, so I suspect that this is the intended path.

JPrevost (changes requested):
Please update the commit message to provide context after squashing / rebasing.

I think links to this form and any interaction with this data can be done in the "process it" ticket.

The comment on naming is not blocking, but the commit message bit is.

matt-bernhardt:
Okay, @JPrevost - I've rebased, squashed, added a real commit message, and renamed the former "file" field to "graduation_list". I hope I caught all the changes - it seems to be okay locally, and the tests pass. Fingers crossed, this is ready for re-review...

JPrevost (approved):

matt-bernhardt:
I've tagged Helen for QA review against the data model, and updated the ticket.

hakbailey:
I agree with Jeremy that we don't currently have a known need for a User field on the registrar data, but it also doesn't hurt and may end up being helpful so might as well leave it. Looks good from the data model perspective!

PR #630, opened 2021-03-02
Fixes implicit submitter role

JPrevost:

* We had a bug where Submitters were having their roles updated in the
  database to `thesis_submitter` which was not the intent and would have
  led to problems in the future.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-193

* This fixes the bug where we set the User role to `thesis_submitter`
  and saved it to the database by instead calling the method
  that ensures they have the appropriate permissions.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -13,9 +13,11 @@ class Ability
         can :manage, :all
       end
 
-      # Promote submitters who do not already have a more elevated role.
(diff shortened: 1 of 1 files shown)


jazairi (approved):
Confirmed locally that making a new Submitter doesn't change that user's role to thesis_submitter. :shipit:

PR #631, opened 2021-03-02
ETD-176 Add relevant audit trail fields to Holds UI

jazairi:

Processors need certain information regarding a Hold's audit trail in the
admin UI:

* The date the hold was released
* The date the thesis file was received
* The user who created the hold

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-176

* Adds new fields to the Hold dashboard: created_by,
dates_thesis_files_received, date_released.
* Adds convenience methods to the Hold model to access the necessary
data.

We haven't yet implemented the primary flag for thesis files, so in this 
iteration, date_thesis_file_received becomes dates_thesis_files_received 
and lists the names and create dates of all associated files.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/dashboards/hold_dashboard.rb
+++ b/app/dashboards/hold_dashboard.rb
@@ -20,7 +20,7 @@ class HoldDashboard < Administrate::BaseDashboard
     date_start: Field::Date,
     date_end: Field::Date,
     date_released: Field::Date,
-    date_thesis_file_received: Field::Date,
+    dates_thesis_files_received: Field::String,
     case_number: Field::String,
     status: Field::Select.with_options(searchable: false, collection: ->(field) { field.resource.class.send(field.attribute.to_s.pluralize).keys }),
     processing_notes: Field::Text,
@@ -58,7 +58,7 @@ class HoldDashboard < Administrate::BaseDashboard
   date_requested
   date_start
   date_end
-  date_thesis_file_received
+  dates_thesis_files_received
   case_number
   status
   processing_notes
--- a/app/dashboards/hold_dashboard.rb
+++ b/app/dashboards/hold_dashboard.rb
@@ -19,9 +19,12 @@ class HoldDashboard < Administrate::BaseDashboard
     date_requested: Field::Date,
     date_start: Field::Date,
     date_end: Field::Date,
+    date_released: Field::Date,
+    date_thesis_file_received: Field::Date,
     case_number: Field::String,
     status: Field::Select.with_options(searchable: false, collection: ->(field) { field.resource.class.send(field.attribute.to_s.pluralize).keys }),
     processing_notes: Field::Text,
+    created_by: Field::Text,
     created_at: Field::Date,
     updated_at: Field::Date,
   }.freeze
@@ -40,6 +43,7 @@ class HoldDashboard < Administrate::BaseDashboard
   status
   date_requested
   date_end
+  date_released
   ].freeze
 
   # SHOW_PAGE_ATTRIBUTES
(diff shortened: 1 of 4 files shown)


app/dashboards/hold_dashboard.rb
```diff
@@ -54,9 +58,11 @@ class HoldDashboard < Administrate::BaseDashboard
   date_requested
   date_start
   date_end
+  dates_thesis_files_received
```
JPrevost, line 61: I think it would be good to include `date_released` on the show page (not sure if that was in the requirements, but it seems useful to include it on this detail page rather than just on the list pages)

app/models/hold.rb
```diff
@@ -38,4 +38,32 @@ def grad_date
   def author_names
     self.thesis.users.map { |u| u.name }.join("; ")
   end
+
+  def created_by
+    if self.versions.first.event == 'create'
+      creator_id = self.versions.first.whodunnit
+      user = User.find_by(id: creator_id)
+      user.kerberos_id
```
JPrevost, line 46: I think it would be good to protect against future staff User deletion here and maybe to a check that `user` exists before and if not displaying something like `User no longer active: user_id XYZ`. [Something more terse would be great but that's the idea of what we likely need to convey for this edge case to prevent error states]

test/models/hold_test.rb
```diff
@@ -89,6 +89,20 @@ class HoldTest < ActiveSupport::TestCase
     assert_equal change.changeset["case_number"], [nil, "2"]
   end
 
+  test 'can list the create dates and names of the parent thesis files' do
+    h = holds(:valid)
+    f = Rails.root.join('test','fixtures','files','a_pdf.pdf')
+    h.thesis.files.attach(io: File.open(f), filename: 'a_pdf.pdf')
+    thesis_file = h.thesis.files.first
+    assert_equal "a_pdf.pdf", thesis_file.filename.to_s
+    assert_equal h.dates_thesis_files_received, "#{thesis_file.created_at.strftime('%Y-%m-%d')} (#{thesis_file.filename.to_s})"
```
JPrevost, line 98: I feel like this is reimplementing the model method in the test and checking it's the same output.

While less convenient, it would probably be better to either explicitly construct what we expect here or look for a pattern that doesn't require reimplementing the method under test.

test/integration/admin_dashboard_test.rb
```diff
@@ -499,6 +499,68 @@ def teardown
...
+                      },
+              }
+    assert_equal orig_count + 1, Hold.count
+
+    hold = Hold.last
+    hold_creator = User.find_by(id: hold.versions.first.whodunnit)
+    assert_equal hold_creator.id, creator.id
+    assert_equal hold_creator.kerberos_id, hold.created_by
+  end
+
```
JPrevost, line 527: We should add a test for `can handle a user who created a hold that no longer exists` [there's a related comment in the `hold.rb` code]

JPrevost (changes requested):
There's a fairly likely edge case around deleted users that may have created Holds at some point I'd like to see us handle as well as a suggestion to consider how we might avoid reimplementing the method we are testing to, er, test.

app/dashboards/hold_dashboard.rb
jazairi: It's not in the requirements, so I'm hesitant to add it now. Would you be comfortable waiting to see if that comes up in stakeholder review?

JPrevost (approved):

jazairi:
As with other Hold UI tickets, stakeholder review will occur in staging.

PR #641, opened 2021-03-10
Update user dashboard to reflect submitters

JPrevost:

* It is difficult for our staff to quickly see which users are
  authorized submitters for various departments
* Making it so looking up the user and being able to quickly see which
  departments they can submit to will allow for initial setup and
  ongoing maintainance more efficient

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-181

* adds Submitters to the User show dashboard
* creates a custom field to jumpstart create a new Submitter

The custom field is forked from the gem's `has_many` code. If there are
changes upstream, we'll need to move them locally to this forked copy.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/controllers/admin/submitters_controller.rb
+++ b/app/controllers/admin/submitters_controller.rb
@@ -42,5 +42,13 @@ module Admin
 
     # See https://administrate-prototype.herokuapp.com/customizing_controller_actions
     # for more information
+    def new
+      resource = new_resource
+      authorize_resource(resource)
+      resource.user_id = params[:user_id]
+      render locals: {
+        page: Administrate::Page::Form.new(dashboard, resource),
+      }
+    end
   end
 end
--- a/app/views/fields/user_submitter_field/_show.html.erb
+++ b/app/views/fields/user_submitter_field/_show.html.erb
@@ -34,5 +34,6 @@ from the associated resource class's dashboard.
   <%= t("administrate.fields.has_many.none", default: "–") %>
 <% end %>
 
-<br />
-<%= link_to("Authorize User to new Department as Submitter", new_admin_submitter_path(user_id: page.resource.id), class: 'button') %>
+<div class='new-submitter-button'>
+  <%= link_to("Authorize User to new Department as Submitter", new_admin_submitter_path(user_id: page.resource.id), class: 'button') %>
+</div>
--- a/app/views/fields/user_submitter_field/_show.html.erb
+++ b/app/views/fields/user_submitter_field/_show.html.erb
@@ -1,0 +1,38 @@
+<%#
+Yanked the entire first section of this code from
+https://github.com/thoughtbot/administrate/blob/master/app/views/fields/has_many/_show.html.erb
+%>
+<%#
+# HasMany Show Partial
+This partial renders a has_many relationship,
+to be displayed on a resource's show page.
+By default, the relationship is rendered
+as a table of the first few associated resources.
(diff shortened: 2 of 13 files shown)


app/controllers/admin/submitters_controller.rb
```diff
@@ -42,5 +42,13 @@ class SubmittersController < Admin::ApplicationController
 
     # See https://administrate-prototype.herokuapp.com/customizing_controller_actions
     # for more information
+    def new
```
jazairi, line 45: Should we add test or two to the admin dashboard controller confirming the behavior in this controller action? (To clarify, that's not a rhetorical question; I'm not sure whether it make sense to test it or not.)

app/views/fields/user_submitter_field/_show.html.erb
```diff
@@ -0,0 +1,38 @@
...
+    table_title: field.name,
+  ) %>
+  <% if field.more_than_limit? %>
+    <%= paginate field.resources(page_number), param_name: "#{field.name}[page]" %>
+  <% end %>
+<% else %>
+  <%= t("administrate.fields.has_many.none", default: "–") %>
+<% end %>
+
+<br />
```
jazairi, line 37: I think we should remove this `br` tag and wrap the 'new submitter' button in a `div` with a top margin of 1rem:
Screen Shot 2021-03-11 at 11 00 12 AM

This would more closely match the spacing when submitters are present:
Screen Shot 2021-03-11 at 11 00 46 AM

jazairi (changes requested):
Custom fields! I'm disproportionately excited about the possibilities this presents.

This looks great, and I'm glad I got to review it because I'll probably steal this approach for the hold history button. My first comment about testing is non-blocking, but I think the styling tweak would be good to implement before stakeholder review.

app/controllers/admin/submitters_controller.rb
JPrevost: yeah good call. I think a case for not passing the param and one passing it and showing it gets selected make sense.

app/views/fields/user_submitter_field/_show.html.erb
JPrevost: But I'm lazy and <br /> fits that well. Also, you are absolutely correct and I'll adjust. Thanks!

JPrevost:
@jazairi I've added the tests you suggested in one commit and a refactor of authorization for administrate in separate commit that was required to actually make the tests work. Please take a close look at these changes as well as confirming with test accounts if you can to ensure the app still works as we expect (I've done this, but need confirmation I'm not missing something).

The CI failing seems to be unrelated, but before we merge I'll dig into what broken in the build (it's before the tests that it fails so hopefully just a blip).

jazairi (approved):
This looks great! Kudos for tackling authorization as part of this. :shipit: 

I did some preliminary testing of the auth updates, and most things work as expected. I noticed one use case from the data model that isn't covered, which is that a thesis_processor should be able to destroy a Thesis. I'm also a little confused about the transfer submitter role, but that's completely unrelated to this ticket. 

I suspect we need a thorough review of the Ability model to make sure all of the roles can do what they're supposed to do. I'm going to repurpose ETD-190, since its original purpose (giving thesis_processors access to administrate) is largely fixed by this PR: https://mitlibraries.atlassian.net/browse/ETD-190

PR #645, opened 2021-03-15
ETD-186 Create UI for full listing of hold history

jazairi:

We need a way to see the full audit trail for a Hold record. Only some
paper_trail fields are currently visible in the Hold administrate UI.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-186

* Adds a hold history page (the Hold show page) that displays audit
information for a given Hold record. Initially, I chose to display the
following fields for each version:
    * The event type (e.g., create/update/destroy);
    * The user who modified the record; and
    * Everything in the version changeset.
* Adds a custom field, hold_history, to administrate, to provide a link
to the hold history page from the Hold dashboard.
* Adds Hold#show method to route to the hold history page.

* Added complexity: we now have a Hold controller with a show method, a
corresponding view and route, and a Hold helper.
* In order to get the custom field to work and display as desired, I
added an i18n translate to render the ID field in the Hold dashboard
as 'Hold history'. We may need to rework this later, if for some reason
we want to display the Hold ID in a different context.
* Because I chose to create this view outside of administrate, it uses
the regular Thing layout. This might be confusing to users since they
will be coming to this view from the admin dashboard, which has a
different layout.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/controllers/hold_controller.rb
+++ b/app/controllers/hold_controller.rb
@@ -1,0 +1,23 @@
+class HoldController < ApplicationController
+  before_action :require_user
+  before_action :authenticate_user!
+  load_and_authorize_resource
+  protect_from_forgery with: :exception
+
+  def show
+    @hold = Hold.find(params[:id])
+  end
+
+  private
+
+  def require_user
+    return if current_user
+    # Do NOT use ENV['FAKE_AUTH_ENABLED'] directly! Use the config. It performs
+    # an additional check to make sure we are not on the production server.
+    if Rails.configuration.fake_auth_enabled
+      redirect_to user_developer_omniauth_authorize_path
+    else
+      redirect_to user_saml_omniauth_authorize_path
+    end
+  end
+end
--- a/app/helpers/hold_helper.rb
+++ b/app/helpers/hold_helper.rb
@@ -1,0 +1,23 @@
+module HoldHelper
+  def modified_by(version)
+    if user = User.find_by(id: version.whodunnit)
+      link_to user.kerberos_id, admin_user_path(user.id)
+    else
+      "User ID #{version.whodunnit} no longer active."
+    end
+  end
+
+  def render_hold_history_field(field_name, field_value)
+    # Set nil values to something that will display onscreen
+    field_value = "n/a" if field_value.nil? 
+
+    # For foreign key fields, return a link to the record
+    if field_name == 'thesis_id' && thesis = Thesis.find_by(id: field_value)
+      field_value = link_to thesis.title, admin_thesis_path(field_value)
+    elsif field_name == 'hold_source_id' && hold_source = HoldSource.find_by(id: field_value)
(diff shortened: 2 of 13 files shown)


app/controllers/hold_controller.rb
```diff
@@ -0,0 +1,23 @@
+class HoldController < ApplicationController
+  before_action :require_user
+  before_action :authenticate_user!
+  load_and_authorize_resource
+  protect_from_forgery with: :exception
+
+  def show
```
JPrevost, line 7: `render layout: false` would allow you to keep this out of our public UI template. (noting this as you mentioned it in the PR info, not because I specifically feel like we should do it).

app/views/hold/show.html.erb
```diff
@@ -0,0 +1,32 @@
...
+    <ul class="list-unbulleted">
+      <li>Event type: <%= version.event %></li>
+      <li>Modified by: <%= modified_by(version) %></li>
+      <li>
+        <table class="table table-simplified table-cozy">
+          <tbody>
+            <tr>
+              <th scope="col">Field name</th>
+              <th scope="col">Previous value</th>
+              <th scope="col">Current value</th>
```
JPrevost, line 17: `Current value` may be misleading when you get into multiple changes to a Hold. Maybe something like `Previous value` and `Changed to value`? Or even `Old` and `New`? 🤷🏻

JPrevost (changes requested):
I'd prefer we change the `Current value` to something else but otherwise 👍🏻

app/helpers/hold_helper.rb
```diff
@@ -11,12 +11,13 @@ def render_hold_history_field(field_name, field_value)
     # Set nil values to something that will display onscreen
     field_value = "n/a" if field_value.nil? 
 
+    # For foreign key fields, return a link to the record
     if field_name == 'thesis_id' && thesis = Thesis.find_by(id: field_value)
```
JPrevost, line 15: field_value=...

app/views/hold/show.html.erb
jazairi: Good catch! It's not really the "current value" unless it's the current version. Change forthcoming...

app/controllers/hold_controller.rb
jazairi: Thanks, that's good to know. I think I'll wait to change anything until Mikki reviews it (largely due to laziness, since using the layout gives me all the CSS I need).

app/helpers/hold_helper.rb
jazairi: I refactored this one slightly. I'm still not crazy about it, but I found myself going down a refactoring rabbit hole that didn't feel especially productive.

JPrevost: I do this this version is clearer than the previous version but I hear you on the rabbit hole.

JPrevost (approved):

PR #668, opened 2021-03-29
ETD-200 Update student submission mailer for new application

jazairi:

We want to send a helpful confirmation email to students when they
submit their thesis info.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-200

This updates the ReceiptMailer to include the information specified in
in the contextual content doc. The mailer is called in the update action
of the Thesis controller.

* In my local environment, I wasn't sure how to  manually test this beyond
looking at the preview. We will want to try sending emails in the PR
build/staging to ensure that the mailer behaves as expected.
* Some weird spacing in the receipt_email fixture was required to pass
the assertion comparing the fixture to the email body.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/mailers/receipt_mailer.rb
+++ b/app/mailers/receipt_mailer.rb
@@ -1,10 +1,10 @@
 class ReceiptMailer < ApplicationMailer
-  def receipt_email(thesis)
-    return if ENV['DISABLE_ALL_EMAIL'] # allows PR builds to disable emails
+  def receipt_email(thesis, user)
+    return unless ENV.fetch('DISABLE_ALL_EMAIL', 'true') == 'false' # allows PR builds to disable emails
+    @user = user
     @thesis = thesis
-    emails = thesis.users.map { |u| u.email }.join(",")
-    mail(to: emails,
-         cc: ENV['THESIS_ADMIN_EMAIL'],
-         subject: 'Thesis Submission Receipt - MIT Libraries')
+    mail(from: "MIT Libraries <#{ENV['THESIS_ADMIN_EMAIL']}>",
+         to: @user.email,
+         subject: 'Your thesis information submission')
   end
 end
--- a/app/views/receipt_mailer/receipt_email.html.erb
+++ b/app/views/receipt_mailer/receipt_email.html.erb
@@ -1,28 +1,45 @@
-<p>Hello,</p>
+<p>Hello <%= @user.given_name %>,</p>
 
-<p>You are receiving this message to confirm your successful submission of
-either your ProQuest form, title page, and abstract or the electronic version of
-your <%= @thesis.graduation_month %> <%=
-@thesis.graduation_year %> thesis "<%= @thesis.title %>" to the MIT Libraries.
-</p>
+<p>You have successfully updated your thesis information. 
+Here is the information we have on record:</p>
 
-<p>If you submitted a May or September 2020 thesis, please note that only the
-ProQuest form, title page, and abstract should have been submitted here. The
-actual thesis should go directly to your department. For degree periods prior to
-May 2020, we accept all electronic copies via this form.</p>
+<ul>
+  <li>Name: <%= @user.display_name %></li>
+  <li>Email: <%= @user.email %></li>
+  <li>ORCID iD: <%= @user.orcid %></li>
+  <li>Thesis title: <%= @thesis.title %></li>
+  <li>Name as it appears on your thesis: <%= @user.preferred_name %></li>
+  <li>Co-author: <%= @thesis.coauthors %></li>
+  <li>Department(s): <%= @thesis.departments.map { |d| d.name_dw }.join(', ') %></li>
+  <li>Degree(s): <%= @thesis.degrees.map { |d| d.name_dw }.join(', ') %></li>
+  <li>Degree date: <%= @thesis.graduation_month %> <%= @thesis.graduation_year %></li>
+  <li>Copyright holder: <%= @thesis.copyright.holder %></li>
+  <li>License: <%= @thesis.license_id ? @thesis.license.display_description : "" %></li>
+  <li>Thesis supervisors: <%= @thesis.advisors.map { |a| a.name }.join(', ') %></li>
+  <li>Abstract: <%= @thesis.abstract %></li>
+  <li>Notes: <%= @thesis.author_note %></li>
+</ul>
 
-<p>You do not need to reply to this message.</p>
+<p>You have until graduation day to update this information.</p>
 
-<p>Theses will be added to DSpace after degrees have been awarded and the thesis
-has been fully processed by the MIT Libraries (typically 3-6 months after the
-degree date).</p>
+<p>A few reminders:</p>
+<ul>
+  <li>Save your files in the correct format.
+    <ul>
+      <li>Thesis PDF: macdonald-mssimon-mcp-dusp-2019-thesis.pdf</li>
+      <li>Proquest Form: macdonald-mssimon-mcp-dusp-2019-proquest.pdf</li>
+      <li>Signature Page: macdonald-mssimon-mcp-dusp-2019-sig.pdf</li>
+    </ul>
+  </li>
+  <li>Submit your thesis files and Proquest form (PhDs only) directly to 
+  your department.</li>
+  <li>Be sure that all of the elements on your title page match the 
+  information you have provided to the Libraries today!</li>
+</ul>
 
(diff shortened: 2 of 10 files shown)


app/mailers/receipt_mailer.rb
```diff
@@ -1,10 +1,10 @@
 class ReceiptMailer < ApplicationMailer
-  def receipt_email(thesis)
-    return if ENV['DISABLE_ALL_EMAIL'] # allows PR builds to disable emails
+  def receipt_email(thesis, user)
+    return unless ENV.fetch('DISABLE_ALL_EMAIL', 'true') == 'false' # allows PR builds to disable emails
```
matt-bernhardt, line 3: TIL the difference between `ENV[]` and `ENV.fetch()` - nice touch :-)

app/views/receipt_mailer/receipt_email.html.erb
```diff
@@ -1,28 +1,44 @@
...
-<p>You do not need to reply to this message.</p>
+<p>You have until graduation day to update this information.</p>
 
-<p>Theses will be added to DSpace after degrees have been awarded and the thesis
-has been fully processed by the MIT Libraries (typically 3-6 months after the
-degree date).</p>
+<p>A few reminders:</p>
+<ul>
+  <li>Save your files in the correct format.</li>
+    <ul>
```
matt-bernhardt, line 28: I'm not sure how finicky HTML rendering is within email, but I don't think this markup is valid - the inner list should be inside a list item element according to the HTML spec. That said, HTML in emails can be weird - so if you did this because of a quirk there, then nevermind me.

test/integration/thesis_test.rb
```diff
@@ -167,6 +167,57 @@ def teardown
...
+    thesis = theses(:two)
+
+    mock_auth(user)
+
+    ClimateControl.modify DISABLE_ALL_EMAIL: 'false' do
+      email = ReceiptMailer.receipt_email(thesis, user)
+
+      assert_emails 1 do
+        patch thesis_path(thesis), params: { thesis: thesis.serializable_hash }
+      end
```
matt-bernhardt, line 181: Something about this block is tripping me up - I think because the `email = ...` on line 177 doesn't appear to relate to the `assert_emails` block from 179-181. However, if I comment it out, then the assertions on 182 & 183 fail.

Could this be slightly re-organized to move 177 down below the assert_email block?  Or is it actually required in some way that I'm missing? I approve of having tests to verify attributes of the email that gets sent, which I think is the intent here - but it feels like things are interleaved here in way that could be avoided.

test/integration/thesis_test.rb
```diff
@@ -167,6 +167,57 @@ def teardown
...
+  end
+
+  test 'a confirmation email is not sent when emails are disabled' do
+    user = users(:basic)
+    thesis = theses(:two)
+
+    mock_auth(user)
+
+    ClimateControl.modify DISABLE_ALL_EMAIL: 'true' do
+      email = ReceiptMailer.receipt_email(thesis, user)
```
matt-bernhardt, line 194: Is this line required?

test/integration/thesis_test.rb
```diff
@@ -167,6 +167,57 @@ def teardown
...
   end
 
+  test 'a confirmation email is sent on form submit' do
+    user = users(:basic)
+    thesis = theses(:two)
+
+    mock_auth(user)
+
+    ClimateControl.modify DISABLE_ALL_EMAIL: 'false' do
+      email = ReceiptMailer.receipt_email(thesis, user)
```
matt-bernhardt, line 177: I may be missing something about the testing environment - does calling this method - which itself calls `mail()` inside it, result in an email being sent due to this line?

test/integration/thesis_test.rb
```diff
@@ -167,6 +167,57 @@ def teardown
...
+    thesis = theses(:two)
+
+    mock_auth(users(:thesis_admin))
+
+    ClimateControl.modify DISABLE_ALL_EMAIL: 'false' do
+      email = ReceiptMailer.receipt_email(thesis, user)
+
+      assert_emails 1 do
+        patch thesis_path(thesis), params: { thesis: thesis.serializable_hash }
+      end
```
matt-bernhardt, line 214: I have the same questions about this test structure as I do for the block from lines 176-185.

test/mailers/receipt_mailer_test.rb
```diff
@@ -1,27 +1,52 @@
...
+      email = ReceiptMailer.receipt_email(thesis, user)
 
-    # Send the email, then test that it got queued
-    assert_emails 1 do
-      email.deliver_now
-    end
+      # Send the email, then test that it got queued
+      assert_emails 1 do
+        email.deliver_now
+      end
```
matt-bernhardt, line 13: This block makes more sense to me, because we're calling a method on the email object - but on the previous file, we're patching a form route, not dealing with the email object.

test/integration/thesis_test.rb
```diff
@@ -167,6 +167,57 @@ def teardown
...
+      assert_not_equal author.email, email.to
+      assert_equal ['thesis_admin@example.com'], email.to
+      assert_equal 'Your thesis information submission', email.subject
+    end
+  end
+
```
matt-bernhardt, line 220: Do we need tests to confirm whether an email is sent from a new thesis form? At the moment the `create` method ends up calling `update`, so the email gets sent that way - but if that logic ever changes, we'll lose the email.

matt-bernhardt (commented):
I've got questions, but in general terms this looks pretty solid. I noted one area where the HTML email template might not be valid, but if this ends up looking okay in practice then I'm fine with ignoring it.

All my other questions are trying to understand the testing logic - and in one case wondering whether an additional test is needed. In general, though, I'm fine with things merging as-is if you feel the current structures are what's needed.

(I also commend you for writing the tests that failed me when I apparently removed the mailer line in previous work... feeling sheepish there)

app/views/receipt_mailer/receipt_email.html.erb
jazairi: Oh, good catch! I was on autopilot while marking that up and missed that one.

test/integration/thesis_test.rb
jazairi: Hm... I actually don't think you're missing anything. This seems to be a case of pernicious copypasta from the receipt mailer test. It looks like the test is passing because it explicitly calls `receipt_email`, not because it does so while updating the thesis record. So, basically it's just doing the same thing as the receipt mailer test.

All that to say that this is broken. Thanks for the callout!

test/integration/thesis_test.rb
jazairi: That's a good question. I think it makes sense as a kindness to our future selves. 👍

test/mailers/receipt_mailer_test.rb
jazairi: That's because this is where I copied that logic from! 😆

test/integration/thesis_test.rb
jazairi: Oh, never mind. On closer look, it _does_ still send the email without that line, so this is less broken than I thought.

test/integration/thesis_test.rb
jazairi: Totally agree. I think moving the `email` variable assignment next to the relevant assertions makes sense, and now that I'm considering it further, we really only need to verify attributes in the test in which we confirm the addressee. (The basic structure of the email is already verified in the receipt mailer tests.)

app/views/receipt_mailer/receipt_email.html.erb
matt-bernhardt: Having now sent emails to myself via gmail and to MIT (which is where I'm guessing most of these emails will be rendered), I see the nesting appears correctly - so I'm less worried about this issue.

matt-bernhardt (approved):
This all looks good now - I'm less concerned about the email markup, and the updated test logic looks good. Thanks for considering these points.

:shipit:

matt-bernhardt (approved):
:shipit:

matt-bernhardt:
Oh, interesting. I'm still fine with this merging, but seeing the way the email looks when fields aren't populated makes me want to confirm that UX is going to look at this from their perspective?

jazairi:
@matt-bernhardt So far Mikki is listed as approver, but I'll ask Stephanie to take a look as well.

PR #676, opened 2021-04-02
ETD-230 Remove degree field from new thesis form

jazairi:

The degree drop-down is difficult to use due to duplicate entries. The
degree will always be supplied by the registrar, so we don’t need it from
creation.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-230

This removes the degree validation from the thesis model and removes the
degree field from the new form.

A confirmation email sent to a student that completes the new form will
show no degree associated with their thesis, which may be confusing.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/thesis.rb
+++ b/app/models/thesis.rb
@@ -63,8 +63,6 @@ class Thesis < ApplicationRecord
 
   validates :departments, presence:
     { message: VALIDATION_MSGS[:departments] }
-  validates :degrees, presence:
-    { message: VALIDATION_MSGS[:degrees] }
 
   validates :files_complete, exclusion: [nil]
   validates :metadata_complete, exclusion: [nil]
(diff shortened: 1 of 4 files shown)


jazairi:
The coverage drop is kind of confusing to me. Assuming I'm reading the coveralls report correctly, these lines in the thesis model are no longer covered:

```
scope :valid_months_only, lambda {
  select { |t| VALID_MONTHS.include? t.grad_date.strftime('%B') }
}
```

But as far as I can tell, this block has nothing to do with the changes I made.

JPrevost (approved):
Interesting note:

If a student is on the `edit` page and not Degree is associated with their thesis (for whatever reason).... it doesn't show up.

If a Degree is associated, it does show up.

As it's a non-editable field and processors will validate everything, this feels fine. Just found it interesting.

jazairi:
@JPrevost I think that's because of how degrees are rendered in the edit form, specifically this block beginning on line 100 with `f.object.degrees.find_all do |deg|`. If I'm understanding that code correctly, it returns an empty array if there aren't any degrees, so it doesn't render anything, which works well in this case!

JPrevost:
@jazairi I agree. This changeset did not introduce the behavior and it feels fine. :shipit:

PR #677, opened 2021-04-02
Allow removal of supervisor fields

matt-bernhardt:
This resolves the final aspect of ETD-240, enabling the user to remove thesis supervisor fields from the form under some conditions. This avoids the problem of a user having created too many fields, as well as allowing user to remove supervisors they have already identified.

At the same time, this change adds a check to the form logic to prevent the user from removing the _last_ supervisor field.

This is all possible because the supervisor fields are not backed by any authority, and all values are specific to each thesis. This work creates a side effect, however, of allowing orphan supervisors to accumulate - because when an existing supervisor is removed from a Thesis, the link is deleted but the supervisor record itself is kept. Eventually we will need a cleanup task, or to convert advisors to something backed by an authority.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/assets/javascripts/student_thesis_form.js
+++ b/app/assets/javascripts/student_thesis_form.js
@@ -3,12 +3,21 @@
 
 // Hide/show license field based on copyright answer
 function conditionalLicenseField() {
-    var value = $("select#thesis_copyright_id option:selected").text();
-    if ('I hold copyright' == value) {
-      $("div.thesis_license").show();
-    } else {
-      $("div.thesis_license").hide();
-      $("select#thesis_license_id")[0].value = "";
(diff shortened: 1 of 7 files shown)


jazairi (approved):
This looks like a reasonable way to solve this problem, and I tested a few use cases and confirmed that it works as expected. :shipit: 

I share your concern about orphaned advisor records. I don't think that needs to be fixed as part of this PR, though. Could you please create a ticket for that work so we can add it to the maintenance epic?

matt-bernhardt:
Sure - the maintenance ticket is https://mitlibraries.atlassian.net/browse/ETD-259 .

PR #697, opened 2021-04-27
First part of Transfer processing

matt-bernhardt:
This implements the first few steps of the Transfer Processing workflow:
- Creates a page showing all transfer records, with links to each transfer detail page
- Implements a jQuery plugin (DataTables) to allow interactive filtering and sorting of the list of Transfers

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/views/transfer/_empty.html.erb
+++ b/app/views/transfer/_empty.html.erb
@@ -1,0 +1,3 @@
+<tr class="empty">
+  <td colspan="6">There were no transfers matching your parameters.</td>
+</tr>
--- a/app/views/transfer/select.html.erb
+++ b/app/views/transfer/select.html.erb
@@ -1,0 +1,63 @@
+<%= content_for(:title, "Transfer Processing | MIT Libraries") %>
+
+<% content_for :additional_js do %>
+  <script src="https://cdn.datatables.net/1.10.24/js/jquery.dataTables.min.js"></script>
+<% end %>
+
+<link href="https://cdn.datatables.net/1.10.24/css/jquery.dataTables.min.css" rel="stylesheet">
+
+<h3 class="title title-page">Transfer Processing Queue</h3>
+
+<div id="term-list" class="filter-row">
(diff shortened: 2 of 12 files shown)


app/views/transfer/select.html.erb
```diff
@@ -0,0 +1,63 @@
+<%= content_for(:title, "Transfer Processing | MIT Libraries") %>
+
+<% content_for :additional_js do %>
+  <script src="https://cdn.datatables.net/1.10.24/js/jquery.dataTables.min.js"></script>
+<% end %>
+
+<link href="https://cdn.datatables.net/1.10.24/css/jquery.dataTables.min.css" rel="stylesheet">
```
JPrevost, line 7: I wonder if our theme gem should allow `<% content_for :additional_css do %>` to better handle this use case?

(No action needed, just thinking out loud).

app/views/transfer/_empty.html.erb
```diff
@@ -0,0 +1,3 @@
+<tr class="empty">
+  <td colspan="6">There were no transfers matching your parameters.
```
JPrevost, line 2: `td` not closed

JPrevost (approved):

app/views/transfer/_empty.html.erb
matt-bernhardt: *smacks head* D'oh! I'm going to push something to fix this.

app/views/transfer/select.html.erb
matt-bernhardt: I was on the point of asking why I couldn't find a layout in this app - didn't think to ask about the theme gem. Yes, I think that would be a good feature to add for a use case like this.

(I don't think that blocks this merging, though right? We can do that via maintenance?)

JPrevost (approved):
Yeah that theme gem thing wasn't intended to be included in this work. :shipit:

PR #708, opened 2021-05-17
Assign selected files to identified Thesis

matt-bernhardt:
This makes the Transfer processing form actually do work, attaching selected files from a Transfer to the Thesis record identified by the user.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/controllers/transfer_controller.rb
+++ b/app/controllers/transfer_controller.rb
@@ -29,13 +29,13 @@ class TransferController < ApplicationController
   def files
     @transfer = Transfer.find(params[:id])
     @thesis = Thesis.find(params[:thesis])
-    flash[:success] = "This is a test of the files method.<br><br>".html_safe
+    flash[:success] = ("The following files have been assigned to '" + @thesis.title + "'<br><br>").html_safe
     filelist = params[:transfer][:file_ids]
     filelist.each do |file|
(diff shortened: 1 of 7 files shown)


app/controllers/transfer_controller.rb
```diff
@@ -29,13 +29,13 @@ def create
   def files
```
JPrevost, line 29: All of the instance variables in this method would be better as locals. They only need to be instance variables when they will be used in the views directly. Since the end of this method redirects to another method, which resets the instance variables, none of these needs to be instance.

Separately, creating so many flash messages makes me wonder if there's another solution, but since we don't use a view for this method and instead redirect away I didn't see a clear alternative. Noting so you knew what was in my brain, but not requesting a change on the use of flashes here.

JPrevost (changes requested):

app/controllers/transfer_controller.rb
matt-bernhardt: I've pushed a commit to convert these variables to locals - thanks for noticing that.

As far as the flash message goes, it may be more fewer lines of code to use a mapping function, rather appending lines inside the loop (for that matter, everything in the loop could be doable via a separate method that receives a map call... something for the future perhaps)

JPrevost (approved):

PR #720, opened 2021-06-10
Bugfix for broken Hold dashboard searches

jazairi:

Certain fields in the Hold dashboard, which are generated via convenience
methods and not stored in the db,  were defined as Field::String. Because
Field::String is searchable by default, this caused the dashboard search
to break when it tried to query nonexistent columns.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETD-304

This changes the offending fields to Field::Text, which are not
searchable by default.

* When I fixed this bug, I noticed that the only searchable field was
status. I added thesis title as another searchable field, since there
aren't many Hold model fields that are searchable in administrate.
* Notably, status is searchable only if you add a colon to the
end of the querystring (e.g., 'active:'), due to some weirdness with
how collection filters work in administrate.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/dashboards/hold_dashboard.rb
+++ b/app/dashboards/hold_dashboard.rb
@@ -8,9 +8,10 @@ class HoldDashboard < Administrate::BaseDashboard
   # which determines how the attribute is displayed
   # on pages throughout the dashboard.
   ATTRIBUTE_TYPES = {
-    thesis: Field::BelongsTo,
-    author_names: Field::String,
-    degrees: Field::String,
+    thesis: Field::BelongsTo.with_options(searchable: true, 
+                                          searchable_fields: ['title']),
+    author_names: Field::Text,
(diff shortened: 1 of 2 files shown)


jazairi:
Hm...locally I'm seeing 100% coverage for the Hold dashboard. Not sure why coveralls is reporting something different.

matt-bernhardt (approved):
Well done!

I'm also not sure why coverage isn't noted for those last two lines in the filter - it seems that they should be counted. When I run the tests locally I'm getting 95.39% coverage, rather than the 94.32% reported here.

:shipit:

jazairi:
Thanks, @matt-bernhardt! I'm seeing 95.39% coverage in the latest run, so I think that coverage report ran before I wrote tests for this. In any case, I think we're good to merge.

PR #725, opened 2021-06-23
Prototype for removing files from a thesis

matt-bernhardt:
#### Developer

#### Code Reviewer

YES | NO

YES | NO


app/controllers/thesis_controller.rb
```diff
@@ -86,6 +94,24 @@ def process_theses_update
 
   private
 
+  def deleted_file_list
+    list = []
+    return list if not thesis_params["files_attachments_attributes"]
+    thesis_params["files_attachments_attributes"].values.select {
```
JPrevost, line 100: Hmmm. I'm not sure what to do with this. We generally prefer `do...end` for multiline blocks but also prefer `{}` for chaining.

I think the line at the end of this style guide rule might be in play here:
```
is this code really readable and can the blocks' contents be extracted into nifty methods?

```

JPrevost: Adding the link might help:
https://github.com/rubocop/ruby-style-guide#single-line-blocks

test/controllers/thesis_controller_test.rb
```diff
@@ -353,4 +353,35 @@ class ThesisControllerTest < ActionDispatch::IntegrationTest
...
+            "0": {
+                "id": th.files.first.id,
+                "_destroy": 1
+              }
+          }
+        }
+      }
+    follow_redirect!
+    # The flash message has a link back to the transfer with the file.
+    assert_select 'div.alert ul li a', text: 'a_pdf.pdf', href: transfer_path(th), count: 1
```
JPrevost, line 383: It's a little unclear what is happening in this test. I feel like it attaches then removes a single file, but this assertion is checking for the presence of the file name which feels confusing (I think it's checking for it in the flash, but if it was still attached I think this would still pass due to it showing up in the record itself?

test/controllers/thesis_controller_test.rb
```diff
@@ -353,4 +353,35 @@ class ThesisControllerTest < ActionDispatch::IntegrationTest
     assert_equal path, thesis_process_path(theses(:one))
   end
 
+  test 'removing a file from a thesis will be reflected with a link to its transfer in the flash message' do
```
JPrevost, line 356: An additional test where multiple files are attached and only one is removed and the rest remain would be helpful I think.

JPrevost (commented):
I've made a few comments / suggestions. I'm happy to discuss them to determine which ones are worth addressing now and which we can just say "yup" and ignore for now.

app/controllers/thesis_controller.rb
matt-bernhardt: Thanks for sharing this link, I probably need to look over the entire style guide a bit more.

I tried reworking this method, looking for ways to improve legibility - can you take a look at the commit that I just pushed up? I'm also happy to jump into a pairing session on this if talking live would help.

test/controllers/thesis_controller_test.rb
matt-bernhardt: I've ended up going with three tests, at this point:
1. Confirm that, when the form is submitted with no deletions, that no files are affected (measured by various counts)
2. Confirm that, when the form is submitted with a deletion, that the counts of records/files/blobs change in the way we expect
3. Confirm that, when the form is submitted with a deletion that the flash message afterwards contains a link back to the transfer record.

Along the way I'm adding a setup method for these three tests, similar to what I did with the transfer controller tests with commit 72bafc8 .

I think this makes this ready for re-review, but if I'm missing a test scenario I'm open to writing more.

JPrevost (approved):

PR #728, opened 2021-06-29
Adds DSpace authority key field to department

matt-bernhardt:
This adds the authority key field to the Department model, in support of future SWORD submissions to DSpace. Field labels are handled via the i18n locales file.

*Please note* the original ticket calls for this field to be editable by thesis processors. We have thus far used the "admin" flag to achieve this ability, so no work is needed on this front. I've done a bit of work to see how processors could be given access to this area without the flag, and - while it is possible - there's some weirdness in how this change would happen that makes me want to set it aside for now. I will set up a separate PR for how that change would happen, to allow us to consider the change de-coupled from the immediate timeline of this project.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-347

#### Developer

#### Code Reviewer

YES

NO

### DIFF
--- a/test/models/department_test.rb
+++ b/test/models/department_test.rb
@@ -37,6 +37,12 @@ class DepartmentTest < ActiveSupport::TestCase
     assert(department.valid?)
   end
 
+  test 'valid without authority key' do
+    department = departments(:one)
+    department.authority_key_dspace = nil
+    assert(department.valid?)
+  end
+
   test 'can have multiple theses' do
     department = departments(:one)
     department.theses = [theses(:one), theses(:two)]
--- a/test/models/department_test.rb
+++ b/test/models/department_test.rb
@@ -2,12 +2,13 @@
 #
 # Table name: departments
 #
-#  id          :integer          not null, primary key
-#  name_dw     :string           not null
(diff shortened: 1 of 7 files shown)


test/models/department_test.rb
```diff
@@ -2,12 +2,13 @@
...
-#  updated_at  :datetime         not null
-#  code_dw     :string           default(""), not null
-#  name_dspace :string
+#  id                   :integer          not null, primary key
+#  name_dw              :string           not null
+#  created_at           :datetime         not null
+#  updated_at           :datetime         not null
+#  code_dw              :string           default(""), not null
+#  name_dspace          :string
+#  authority_key_dspace :string
```
JPrevost, line 11: To match the rest of this test suite, I think a `valid without authority_key_dspace` test would be good.

JPrevost (changes requested):

test/models/department_test.rb
matt-bernhardt: Yeah, good call - I just pushed a commit to add this test.

JPrevost (approved):

PR #730, opened 2021-07-01
Adds searchable flag to certain transfer fields

matt-bernhardt:
This enables staff searching of records within the admin transfer panel.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-334

One caveat is that this will make the graduation year searchable, but not the graduation month. Additionally, searching much match against one and only one field - so something like "SHASS Joe" will fail, even if both "SHASS" and "Joe" would each return an overlapping set of records.

Another caveat is that I haven't asked for stakeholder review of this change yet - without a robust set of records in a review app, I'm not sure how useful that would be. I'm thinking that we can take care of this after the code merges and it gets to the staging dataset.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/dashboards/transfer_dashboard.rb
+++ b/app/dashboards/transfer_dashboard.rb
@@ -8,11 +8,18 @@ class TransferDashboard < Administrate::BaseDashboard
   # which determines how the attribute is displayed
   # on pages throughout the dashboard.
   ATTRIBUTE_TYPES = {
-    user: Field::BelongsTo,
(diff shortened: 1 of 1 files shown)


jazairi (approved):
Confirmed that the fields intended to be searchable are now searchable. Also good to know that it's not possible to search multiple fields in the same query. I hadn't noticed that, so thanks for figuring it out and adding it to the commit message!

PR #740, opened 2021-07-14
Only display unassigned files by default

JPrevost:

* The list of files can be quite large in some cases
* Scrolling through files that have already been processed can be
  annoying and inefficeint
* Most of the time, it is not useful to see the already assigned files

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-366

* Enables conditional display of the already assigned files during
  the Transfer processing workflow
* The default is to not display assigned files

* Choosing to display assigned files should be sticking throughout the
  processing of that Transfer but would need to be reapplied when
  changing to a different Transfer. Being sticky beyond what we have
  implemented here would require using a session or local storage which
  didn't seem necessary at this time but wouldn't be difficult to
  implemente in the future if that is the desired behavior.
* The display of the `<li>` elements had to be adjusted slightly to
  prevent blank `<li>`s being displayed for the files we are skipping.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/controllers/transfer_controller.rb
+++ b/app/controllers/transfer_controller.rb
@@ -34,7 +34,6 @@ class TransferController < ApplicationController
 
       ReceiptMailer.virus_detected_email(@transfer).deliver_later
       redirect_to transfer_confirm_path
-
   end
 
   def files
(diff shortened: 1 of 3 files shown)


matt-bernhardt (approved):
Huh... coming into this ticket I was expecting it to go one of two ways (filter the list via javascript without a page load, or split the record loading in the controller). This is going down a third path, but I think I see the motivation. Interesting. This both allows for testing of the feature (harder with a javascript implementation), and doesn't grow the controller as much as I thought it might (with multiple record loading paths).

The tests pass, I see the feature working, so :shipit:

PR #766, opened 2021-08-13
Filter list of theses on file selection during thesis processing

matt-bernhardt:
This adds a class to the checkboxes in the list of files on a transfer processing form, along with some javascript that detects clicks on those checkboxes. The javascript will add a search/filter value to the table of thesis records, provided the table is currently unfiltered.

If the table is already filtered (either by this feature or via manual entry), no action is taken. We have no way of determining whether any filtering was done by the user or not, so it is better that we leave it alone - and let the user manipulate the filter value themselves.

#### Ticket:

- https://mitlibraries.atlassian.net/browse/ETD-367

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/views/transfer/show.html.erb
+++ b/app/views/transfer/show.html.erb
@@ -65,7 +65,7 @@
         <% if file.blob.attachments.where('record_type = ?', "Thesis").count == 0 %>
         <li>
           <label>
-            <%= check_box_tag( "transfer[file_ids][]", file.id, false, data: { "msg" => "Required - please select at least one file to transfer." }, class: "required" ) %>
+            <%= check_box_tag( "transfer[file_ids][]", file.id, false, data: { "msg" => "Required - please select at least one file to transfer." }, class: "required checkbox-trigger-filter" ) %>
             <%= link_to rails_blob_path(file, disposition: 'inline'), target: :_blank do %>
               <%= file.filename.to_s %>
             <% end %>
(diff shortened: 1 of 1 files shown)


app/views/transfer/show.html.erb
```diff
@@ -107,11 +107,36 @@
...
+        if (debug) { console.log('this was checked') };
+        var filename = $(this).next().text();
+        if (filename.split('-').length > 1) {
+          var kerb = filename.split('-')[1]
+          if (debug) { console.log('Kerb: ' + kerb) };
+
+          if (document.getElementById('thesisTargets_filter').getElementsByTagName('input')[0].value=='') {
+            if (debug) { console.log('Filter field is empty, so we populate') };
+            // document.getElementById('thesisTargets_filter').getElementsByTagName('input')[0].value=kerb;
+            table.search( kerb ).draw();
```
JPrevost, line 125: Oh this sets the value and searches? Nice!

matt-bernhardt:
The debugging feature has now been removed. I also wanted to leave a note on a possible alternative logic which we discussed and decided not to pursue for now (this is also stated in the Jira ticket).

If we wanted to clear the search field via this approach, an `else` clause on the `this.checked` structure should detect unchecks. Within this clause, if we compare the extracted kerb to the search value, we could probably identify cases where clearing the search field is appropriate (i.e. the unchecked value exactly matches the current search).

jazairi (approved):
Looks good! I confirmed that when the text box is empty, if I click a file, it populates the search box with the intended value and filters accordingly. My understanding from the comments in the ticket is that it's okay that this will not clear/overwrite the content of the search field. I think we're :shipit:.

PR #767, opened 2021-08-17
Fix sorting and filtering bugs in DataTables implementations

matt-bernhardt:
This addresses a few different bugs that have been noticed in our DataTables implementation:
1. Both the Thesis and Transfer queues have a problem in how we've implemented the custom filtering buttons. The buttons do not persist their behavior beyond that click action, so re-sorting the table resets the filtering.
2. The Thesis queue, specifically, does not sort by degree date accurately - it uses an alphabetical sort rather than a date sort.
3. The "no records" display on the thesis queue has the wrong number of columns - we never bumped the colspan attribute when we added the issues found column.

Each of these is fixed in a separate commit, because they are separate bugs. I didn't create a separate ticket for the second and third items.

This code review is happening prior to stakeholder review - if I'm forgetting the order here, please let me know and I'll get their input first.

#### Ticket

https://mitlibraries.atlassian.net/browse/ETD-386

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/views/thesis/_select_empty.html.erb
+++ b/app/views/thesis/_select_empty.html.erb
@@ -1,3 +1,3 @@
 <tr class="empty">
-  <td colspan="5">No theses found</td>
+  <td colspan="6">No theses found</td>
 </tr>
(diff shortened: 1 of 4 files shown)


jazairi (approved):
I'm new to DataTables so the last commit feels a little magical to me, but I think it makes sense and it's working as expected. :shipit:

PR #778, opened 2021-08-30
Initial reporting dashboard

matt-bernhardt:
This creates the first summary dashboard for thesis records (for a given term). It contains the data points asked for in ETD-412, with links to see more information for various subsets of records.

1. The total number of thesis records (with a link to see them in the admin dashboard)
2. The number of theses which have attached files (and a link to see them in the processing queue)
3. The number of theses which have been flagged as having issues (there is no link for this box, because there's no easy web address to surface them - you would need to enter a filter value in the processing queue)

There are also two breakdown tables, showing how many thesis records are in each of two categories: publication status and copyright holder. The ticket only asks for publication status, but a separate ticket (413) is going to ask for copyright, and I needed to make sure that the engineering of the table code was at least somewhat abstracted.

As part of this work, the site navigation has also been restructured, removing the duplicate thesis report in favor of a "Report" item which links to this dashboard. Both the dashboard and the duplicate report have a reporting submenu. Further tickets will add more report options to this menu.

Beyond this initial implementation, the thesis controller has been refactored due to feedback from Codeclimate. Two helper methods have been defined in order to reduce the complexity of three methods which have _very_ similar logic. One method extracts a list of terms from a set of thesis (which is then used in a shared filtering partial). The other method filters a set of theses by the `graduation` parameter in the querystring.

I've tried to structure this PR as clearly as I can - but I've also spent longer on this than I intended, and I'm not sure more time will help. I fully expect that QA testing may surface requests for changes, and there are definitely tickets coming that will build upon this work.
 
#### Ticket

https://mitlibraries.atlassian.net/browse/ETD-412

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -1,0 +1,92 @@
+class Report
+  delegate :url_helpers, to: 'Rails.application.routes'
+
+  def card_files(collection, term)
+    subset = collection.joins(:files_attachments)
+    {
+      'value' => subset.pluck(:id).uniq.count,
+      'label' => 'have files attached',
+      'note' => 'Only theses with a status of "Not ready for publication" and "Publication review" will be visible '\
+                'in the processing queue.',
+      'link' => {
+        'url' => url_helpers.thesis_select_path(graduation: term),
+        'text' => "See #{subset.where('publication_status != ?', 'Published').pluck(:id).uniq.count} unpublished "\
+                  'theses in processing queue'
+      }
+    }
+  end
+
+  def card_issues(collection)
+    {
+      'value' => collection.where('issues_found = ?', true).count,
+      'label' => 'flagged with issues'
+    }
+  end
+
+  def card_overall(collection, term)
+    {
+      'value' => collection.count,
+      'label' => 'thesis records',
+      'link' => {
+        'url' => url_helpers.admin_theses_path(search: term),
+        'text' => 'See all in admin dashboard'
+      }
+    }
+  end
+
+  def dashboard_data(collection, term)
+    result = {}
+    result['overall'] = card_overall collection, term
+    result['files'] = card_files collection, term
+    result['issues'] = card_issues collection
+    result
+  end
+
+  def dashboard_tables(collection)
+    result = {}
+    result['status'] = table_status collection
+    result['copyright'] = table_copyright collection
+    result
+  end
+
+  def extract_terms(collection)
+    collection.pluck(:grad_date).uniq.sort
+  end
+
+  def table_populate_defaults(data, values)
+    values.each do |value|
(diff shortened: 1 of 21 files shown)


matt-bernhardt:
@JPrevost and @jazairi, can I ask whether you'd prefer that I squash this work - particularly the second phase - before asking for code review?

The first phase of this work is in the first four commits (finishing with commit 49a3b76 ), which builds out a report dashboard under the Thesis controller. After some initial feedback from Jeremy, I think started to refactor things to run out of a new Report controller and model. The last 15 commits in this PR are that work happening.

I've done this squash locally, but ran into an issue force-pushing the result using Github Desktop. This is resolvable, force-push from the command line, but I'm nervous about whether the refactoring should be "more than one commit". Hence my question about whether I should squash first.

To be clear, I'm not asking for code review _quite_ yet - but whether the branch as you see it here is too messy for code review to happen.

app/models/report.rb
```diff
@@ -0,0 +1,92 @@
+class Report
+  delegate :url_helpers, :to => 'Rails.application.routes'
```
JPrevost, line 2: Code Climate/Rubocop have an appropriate suggestion to use the 1.9+ Hash syntax here.

app/views/thesis/deduplicate.html.erb
```diff
@@ -6,40 +6,32 @@
...
-      <th scope="col">Coauthor(s)</th>
-      <th scope="col">Department(s)</th>
-      <th scope="col">Degree date</th>
-    </tr>
-  </thead>
-  <tbody>
-    <%= render(partial: 'thesis/duplicate_thesis', collection: @thesis) || render('select_empty') %>
-  </tbody>
-</table>
+<div class="layout-3q1q layout-band">
```
JPrevost, line 9: We may want to consider moving the report menu (eventually) to allow for full width reports. For now this is fine but want us to be considering other options since we have unfortunately set a max width in our core styles so there is no way to get a "wide view" of these reports and taking up a portion with a menu is unfortunate.

JPrevost (approved):

app/views/thesis/deduplicate.html.erb
matt-bernhardt: Yeah, I agree that this is unfortunate. I didn't feel like this was the place to solve a fly-out submenu (which is the only alternative I can think of at the moment, aside from a wholesale design system update).

If you like, I can make a ticket in the backlog to revisit the submenu layout with some UX assistance.

matt-bernhardt: I've made ticket ETD-442 for consideration of this question.

PR #789, opened 2021-09-22
Adds a Departments report

matt-bernhardt:
This adds a new report page, focusing on departments information.

#### Ticket

https://mitlibraries.atlassian.net/browse/ETD-417

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/test/controllers/report_controller_test.rb
+++ b/test/controllers/report_controller_test.rb
@@ -138,6 +138,76 @@ class ReportControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
+  # ~~~~~~~~~~~ Departments report ~~~~~~~~~~~~~~~~~~~~~~~~~~
+  test 'departments report exists' do
+    sign_in users(:admin)
+    get report_departments_path
+    assert_response :success
+  end
+
+  test 'anonymous users are prompted to log in by departments report' do
+    # Note that nobody is signed in.
+    get report_departments_path
+    assert_response :redirect
+  end
+
+  test 'basic users cannot see departments report' do
+    sign_in users(:basic)
+    get report_departments_path
+    assert_redirected_to '/'
+    follow_redirect!
+    assert_select 'div.alert', text: 'Not authorized.', count: 1
+  end
+
+  test 'submitters cannot see departments report' do
+    sign_in users(:transfer_submitter)
+    get report_departments_path
+    assert_redirected_to '/'
+    follow_redirect!
+    assert_select 'div.alert', text: 'Not authorized.', count: 1
+  end
+
+  test 'processors can see departments report' do
+    sign_in users(:processor)
+    get report_departments_path
+    assert_response :success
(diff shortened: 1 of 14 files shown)


test/controllers/report_controller_test.rb
```diff
@@ -67,4 +68,74 @@ class ReportControllerTest < ActionDispatch::IntegrationTest
...
+    assert_select '.card-departments span', text: '1 department(s) have transferred files', count: 1
+    assert_select '.list-no-transfers li', count: 2
+  end
+
+  test 'departments report includes instance variables' do
+    sign_in users(:processor)
+    get report_departments_path
+    assert_not_nil assigns(:terms)
+    assert_not_nil assigns(:data)
+    assert_not_nil assigns(:list)
```
JPrevost, line 131: This test seems to require adding a support gem for `assigns`. I'm not sure how much value we get out of testing that instance variables are being set like this. Do you have any thoughts on whether this is testing anything specifically useful versus just covering the lines in code coverage?

I haven't looked closely at the rest of the pr but that new dependency stuck out so I wanted to discuss that up front before digging in further.

matt-bernhardt: Yeah, this is a good question. The new gem in this PR is what enables this test, and that is the only use we're getting from it.

I wanted to write tests like this to be able to understand which instance variables are being sent out by the controller, as a way of confirming that we're not losing components of the dashboard as things grow. I'm open to hearing that either there's a better way to do this, or that this sort of test isn't needed. Part of the value I see in this approach is that, particularly while we're building out the reporting suite and code may change significantly, having a way to check that we've removed this or that component of a report can be valuable (something being borked is easier to detect than when one out of 5 cards on a page disappears).

JPrevost: Gotcha.

You can access instance variables within the test like this to accomplish the same thing without adding that dependency:

```ruby
    assert_nil(controller.instance_variable_get(:@popcorn)) # just showing that this works as expected
    assert_not_nil(controller.instance_variable_get(:@terms))
    assert_not_nil(controller.instance_variable_get(:@data))
    assert_not_nil(controller.instance_variable_get(:@list))
```

matt-bernhardt: I've removed that gem, and re-written the tests to use this syntax. I've chosen not to test for the absence of any instance variable, unless there's a way to use `instance_variable_get` to get the full list of variables - checking for an arbitrary string not existing doesn't feel useful to me since there isn't a variable we want to make sure is excluded.

matt-bernhardt:
Closing this PR. Discussions about reporting have led us to shelve this work for now. It may be revived in the spring, should the need for this report continue. For now, the thesis report has a breakdown of departments, which give us a very similar perspective.

Should the transfer processing work lag, or there be a need to specifically break down _Transfers_ by department (rather than Thesis records by department), we can always revive or revisit this.

PR #790, opened 2021-09-22
Adds tabular displays for license, hold sources, and departments

matt-bernhardt:
This restructures the reporting area slightly, and may allow us to (soon) remove the original page that we added.

The default display at `/report` will now be a view across terms, showing a set of tables of various information about thesis record across the app. Column headings will be links to see more information about that term.

Following any of those links will take the user to `/report/term`, which is the current page for information from a single term. Once on the term-specific view, the user can jump to different terms (or to the all-terms totals) via the dropdown filter (same as now). There is also a link to bring the user back to the `/report` dashboard.

The current differences between the tabular display and the single term display are:
* The tabular display includes a row counting the total number of files which have been attached to theses. This row was discussed on Jira, and I'm leaving it in place for now.
* The single term display has some notes, and links off to other displays to inspect more closely.
* The single term display has a breakdown of thesis holds, which I've yet to port to the overall dashboard.
* The sorting order of some categories on the single term display is different than the tabular display, because of how the default values are populated. Where they differ, the tabular display is probably more consistent with the rest of the application, and is preferable.
* Visually, the tabular display is more consistent in how it treats all values. I'm not sure if all these values are of equal importance, however.

Generally, I think I'd like to have a conversation with Jess and Mikki after this merges, as part of the QA review, about these type of differences, and whether we even need a single term display after the tabular dashboard arrives.

#### Developer

#### Code Reviewer

NO

NO


test/models/report_test.rb
```diff
@@ -8,4 +8,31 @@ class ReportTest < ActiveSupport::TestCase
...
+  test 'pad_category will add rows for each expected value in the category' do
+    r = Report.new
+    expected_rows = Thesis.publication_statuses.count
+    returned_rows = r.index_data['Publication status'].count
+    # Not all licenses are represented in our fixtures...
+    assert_not_equal expected_rows, Thesis.all.joins(:license).pluck(:display_description).uniq.count
+    # ...but by the time we get data back for this category, all rows are present.
+    assert_equal expected_rows, returned_rows
+    # This includes a row of all zeros, which could only have been generated by pad_category
+    assert_equal r.index_data['License'][3][:data].values, [0, 0, 0, 0, 0, 0, 0]
```
JPrevost, line 23: Does this padding always result in all zeroes?

test/models/report_test.rb
```diff
@@ -8,4 +8,31 @@ class ReportTest < ActiveSupport::TestCase
...
+  test 'pad_row will add zeros for undefined academic terms' do
+    r = Report.new
+    expected_terms = Thesis.pluck(:grad_date).uniq.count
+    returned_columns = r.index_data['Summary'][3][:data].count
+    # Not all terms have theses with issues in our test fixtures...
+    assert_not_equal expected_terms, Thesis.all.group(:grad_date).where('issues_found = ?', true).count.count
+    # ...but by the time we get data for display back from a report, all columns are present.
+    assert_equal expected_terms, returned_columns
+    # Test for a row of expected values
+    assert_equal r.index_data['Summary'][3][:data].values, [0, 0, 0, 0, 0, 0, 0]
```
JPrevost, line 36: Does this always result in all zeroes?

app/models/report.rb
```diff
@@ -118,4 +262,111 @@ def table_copyright(collection)
...
+    # This builds out the initial set of a category of rows, based on a provided array (like
+    # Thesis.publication_statuses).
+    output = {}
+    array.each do |item|
+      output[item] = {}
+    end
+    output
+  end
+
+  def pad_row(data)
```
JPrevost, line 364: This feels like it is doing more than padding and is actually populating rows with data? (possibly unlike `pad_category`?)

app/models/report.rb
```diff
@@ -60,46 +60,190 @@ def card_overall(collection, term)
...
+  def data_issues
+    {
+      label: 'Flagged with issues',
+      data: pad_row(Thesis.all.group(:grad_date).where('issues_found = ?', true).count)
+    }
+  end
+
+  def data_multiple_authors
+    row_data = {}
+    sql = 'SELECT t.grad_date, count(t.id) as pop '\
```
JPrevost, line 139: Did you consider heredoc or %Q syntax for multiline strings?

matt-bernhardt: Having looked again at these syntaxes, I think Heredoc makes the most sense. I like its compactness compared with the blank lines needed by %Q (especially given how long these methods are already). Thanks for the suggestion! This will be addressed by the next commit (underway now).

app/models/report.rb
```diff
@@ -60,46 +60,198 @@ def card_overall(collection, term)
...
+  def data_issues
+    {
+      label: 'Flagged with issues',
+      data: pad_terms(Thesis.all.group(:grad_date).where('issues_found = ?', true).count)
+    }
+  end
+
+  def data_multiple_authors
+    row_data = {}
+    query = <<~SQL
```
JPrevost, line 139: Oh those heredocs ended up even cleaner than I was hoping. Nice!

app/models/report.rb
```diff
@@ -118,4 +270,129 @@ def table_copyright(collection)
...
+    #  Thu, 01 Sep 2022=>2}
+    output = {}
+    Thesis.pluck(:grad_date).uniq.sort.each do |term|
+      output[term] = 0
+      output[term] = data[term] if data.include? term
+    end
+    output
+  end
+
+  def populate_category(array)
```
JPrevost, line 385: Much clearer name on this and pad_terms! Thanks!

JPrevost (approved):

PR #791, opened 2021-09-22
ETD-394 Upload DSS metadata JSON files to S3

jazairi:

This is additional work toward ETD-394, which involves creating
a DSS-compliant JSON package and submitting a message to the SQS
queue. This particular commit associates metadata JSON objects
with Thesis records, and puts the metadata files in place in S3.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETD-394

This uses Active Storage to attach metadata JSON (i.e., DspaceMetadata
objects) to their Thesis records and upload them to S3. The next
step is to build the SQS message and the job that puts it all together
and submits it to the SQS queue.

We're using StringIO to simulate IO on the serialized JSON string, so
Active Storage will attach it as a file. This doesn't affect the
outcome, but it's worth noting as a fairly unusual approach.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/test/models/thesis_test.rb
+++ b/test/models/thesis_test.rb
@@ -769,4 +769,28 @@ class ThesisTest < ActiveSupport::TestCase
     new_thesis.update(title: 'updated')
     refute new_thesis.new_thesis?
   end
+
+  test 'supports one DSpace metadata attachment per thesis' do
+    thesis = theses(:one)
+
+    # Add metadata required for DspaceMetadata class
+    file = Rails.root.join('test', 'fixtures', 'files', 'a_pdf.pdf')
+    thesis.files.attach(io: File.open(file), filename: 'a_pdf.pdf')
+    thesis.files.first.description = 'My thesis'
+    thesis.files.first.purpose = 'thesis_pdf'
+    thesis.save
+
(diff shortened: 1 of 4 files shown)


test/models/thesis_test.rb
```diff
@@ -765,4 +765,31 @@ class ThesisTest < ActiveSupport::TestCase
...
     refute new_thesis.new_thesis?
   end
+
+  test 'supports one DSpace metadata attachment per thesis' do
+    thesis = theses(:one)
+
+    # Add metadata required for DspaceMetadata class
+    degree = thesis.degrees.first
+    degree.degree_type_id = degree_types(:bachelor).id
+    degree.save
```
JPrevost, line 775: I'm not sure why we need this part? Does the fixture for the Degree that this Thesis uses not using a valid Degree Type? And if so, does fixing that break a lot of other tests?

JPrevost (approved):

test/models/thesis_test.rb
jazairi: Good point. All tests pass if we associate the bachelor degree type with the first degree fixture, so I think it's fine (and preferable) to do it that way. I pushed up a commit updating that here and in the DspaceMetadata tests.

JPrevost (approved):

PR #792, opened 2021-09-24
ETD-394 Construct DSS SQS message

jazairi:

This is the third part of ETD-394, supporting publication to
DSpace@MIT via the DSpace Submission Service.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETD-394

This constructs the SQS message attributes and body, as per the
DSS spec. Submitting the message to the queue will come in a
subsequent PR.

None.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/sqs_message.rb
+++ b/app/models/sqs_message.rb
@@ -1,0 +1,59 @@
+# Generates an SQS message containing the data needed for the DSpace Submission Service (DSS) to publish a Thesis object
+# to DSpace@MIT. This class assumes that the input Thesis is ready for publication and has an attached DspaceMetadata
+# object.
+
+class SqsMessage
+  def initialize(thesis)
+    @thesis = thesis
+    @package_id = "etd_#{@thesis.id}"
+    @metadata_uri = thesis.dspace_metadata.blob.url
+  end
+
+  def message_attributes
+    attributes = {}
+    attributes['PackageID'] = { 'DataType' => 'String', 'StringValue' => @package_id }
+    attributes['SubmissionSource'] = { 'DataType' => 'String', 'StringValue' => 'ETD' }
+    attributes['OutputQueue'] = { 'DataType' => 'String', 'StringValue' => ENV['OUTPUT_QUEUE_NAME'].to_s }
+    attributes
+  end
+
+  def message_body
+    body = {}
+    body['SubmissionSystem'] = 'DSpace@MIT'
+    body['CollectionHandle'] = collection_handle
+    body['MetadataLocation'] = @metadata_uri
+    body['Files'] = map_files
+
+    # SQS requires the MessageBody to be a string
+    body.to_json
+  end
+
+  def map_files
+    @thesis.files.map do |f|
+      {
+        'BitstreamName' => f.blob.filename.to_s,
+        'FileLocation' => f.blob.url,
+        'BitstreamDescription' => bitstream_description(f)
+      }
+    end
+  end
+
+  # There is a handle for all MIT theses, but there are also subcollections for doctoral, graduate, and undergraduate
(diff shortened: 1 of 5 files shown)


jazairi:
@ehanson8 In case you're curious, here's my approach to building the DSS submission message.

app/models/sqs_message.rb
```diff
@@ -0,0 +1,48 @@
...
+    # SQS requires the MessageBody to be a string
+    body.to_json
+  end
+
+  def map_files
+    @thesis.files.map do |f|
+      {
+        'BitstreamName' => f.blob.filename.to_s, # is this also the filename in DSpace?
+        'FileLocation' => f.blob.url,
+        'BitstreamDescription' => -> { f.description ? [f.purpose, f.description].join(' ') : f.purpose }.call
```
jazairi, line 32: Note to reviewer: using a lambda here seemed like the most sensible way to deal with a Rubocop style issue. The alternative was string interpolation, which Rubocop generally doesn't like in this sort of situation (and I tend to agree). I'm happy to revert to interpolation if we think this is significantly less readable.

JPrevost: I don't see rubocop complaining about this syntax that I think does the same thing?

```
'BitstreamDescription' => "#{f.purpose} #{f.description}".strip
```

JPrevost: Also, are we sure we want to put the raw purpose in here? `thesis_pdf some description` seems weird to me (we can always adjust later if we weren't given a lookup table for what to actually put in here)

JPrevost (changes requested):
Please take a look at my suggestion around how to eliminate the lambda. If I'm missing something important, we can approve this as-is but I think if the alternate approach works it's much easier to read (tests pass so It think it's okay-ish?)

app/models/sqs_message.rb
jazairi: Ha! Yeah, you're right. Now I want to know why Rubocop was flagging that approach for me last week...must've been a typo.

As for the purpose and description, my reading of the spec is that the bitstream description contains at minimum the file purpose, with the description concatenated if present. I got that from lines 17-18 in the mapping spreadsheet: https://docs.google.com/spreadsheets/d/1JhNvD_0VXCacQRtRgWN3RUX6-OOeTJKv0AbnZFzjgHw/edit#gid=17335869. I think there are still crosswalk issues to resolve with those fields, though, so that may change later.

app/models/sqs_message.rb
```diff
@@ -0,0 +1,48 @@
+class SqsMessage
+  def initialize(thesis)
+    @thesis = thesis
+    @package_id = "etd_#{@thesis.id}"
+    @metadata_uri = thesis.dspace_metadata.blob.url
```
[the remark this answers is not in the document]
jazairi: That's an excellent point, and it reminded me that I never gave context for the `DspaceMetadata` class, either. I'll add a comment to that as well.

app/models/sqs_message.rb
JPrevost: Yeah I think what we have here is fine, but the literal wording "thesis_pdf" seems suspiciously not super user friendly and we'll probably want someone to map us to preferred wording eventually. Maybe a ticket to confirm wording on the options here would be good or we can just ask stakeholders to confirm it when we send samples into test dspace later on.

JPrevost (approved):

app/models/sqs_message.rb
jazairi: Ah, now I understand what you're saying. So maybe a convenience method that translates the raw purposes into a more readable format?

JPrevost (approved):

PR #795, opened 2021-09-28
Adds report for thesis files with no defined purpose

matt-bernhardt:
This adds a new report page for files with no defined purpose. The table includes links to the processing form for that thesis, so staff can assign a purpose.

More details in the commit message...

#### Ticket

https://mitlibraries.atlassian.net/browse/ETD-418

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/controllers/report_controller.rb
+++ b/app/controllers/report_controller.rb
@@ -6,6 +6,14 @@ class ReportController < ApplicationController
 
   include ThesisHelper
 
+  def files
+    report = Report.new
+    theses = Thesis.all
+    @terms = report.extract_terms theses
+    subset = filter_theses_by_term theses
+    @list = report.list_unattached_files subset
+  end
+
   def index
     report = Report.new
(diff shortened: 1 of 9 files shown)


app/controllers/report_controller.rb
```diff
@@ -6,6 +6,14 @@ class ReportController < ApplicationController
 
   include ThesisHelper
 
+  def files
```
JPrevost, line 9: We may want to consider a more explicit name for this. In the future looking at this method it won't be clear that this is really for the `files_with_no_defined_purpose` versus some generic `files` route without having to poke around further.

JPrevost (approved):

app/controllers/report_controller.rb
matt-bernhardt: This is fair, but I'd like to move this to a different name once we have an idea of what the alternative report is going to be. Looking at the current reporting epic in Jira, no other requested report seems likely to overlap with this method.

PR #799, opened 2021-10-05
Adds "Publish to DSpace@MIT" button to processing queue

matt-bernhardt:
This augments the Thesis Processing Queue, adding a button that will publish records in the `Publication Review` status to our repository via the coming DSpace Submission Service.

As yet, this button does not do anything - but it will. Clicking the button with this PR will generate a flash message that is a placeholder.

I've populated a sample record in the review app, in case you want to see how the button appears / disappears.

#### Ticket

https://mitlibraries.atlassian.net/browse/ETD-310

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/controllers/thesis_controller.rb
+++ b/app/controllers/thesis_controller.rb
@@ -44,6 +44,23 @@ class ThesisController < ApplicationController
     @thesis.association(:advisors).add_to_target(Advisor.new) if @thesis.advisors.count.zero?
   end
 
+  def publish_to_dspace
+    # This will be where the publication job gets invoked.
+    # subset = filter_theses_by_term Thesis.in_review
+    # result = subset.map(&:some_async_method)
+    # if result.map { |val| val == true }.reduce(:&)
+    #   flash[:success] = 'This would have published theses to DSpace@MIT.'
+    # else
+    #   flash[:warning] = 'Not all selected theses are ready for publication.<br>'.html_safe + ready.to_s
+    # end
+    if params[:graduation] == 'all' || params[:graduation].nil?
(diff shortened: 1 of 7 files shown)


app/controllers/thesis_controller.rb
```diff
@@ -44,6 +44,19 @@ def edit
     @thesis.association(:advisors).add_to_target(Advisor.new) if @thesis.advisors.count.zero?
   end
 
+  def publish_to_dspace
+    # This will be where the publication job gets invoked.
```
JPrevost, line 48: This made me realize the publishing job currently publishes one thesis, but the expectations here are that we are publishing a specific set. While we could loop over the set here and spin up 1000 jobs, we might also considering refactoring the publishing job to do that work as I'm worried spinning up 1000 jobs might take longer than the heroku timeout and itself need to be backgrounded. <sigh>

app/views/thesis/select.html.erb
```diff
@@ -33,6 +33,25 @@
   </tbody>
 </table>
 
+<div id="publish-to-dspace" class="gridband" style="display: none;" aria-live="polite" role="region">
+  <div class="inline-action well">
+    <% if params[:graduation] == 'all' || params[:graduation].nil? %>
```
JPrevost, line 38: I think, unlike almost all other similar parts of this app, we want the default to not be `all` here but an empty set. We should only publish one term at a time.

JPrevost: We'll need to be careful to ensure the `publish_to_dspace` method checks for a degree param as well I think?

PR #800, opened 2021-10-07
ETD-394 Send messages to SQS queue for DSpace publication

jazairi:

In order to publish theses to DSpace, we need to send messages to
an SQS queue that the DSpace Submission Service (DSS) can read from
and continue the publication workflow.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-394

This creates a DspacePublicationJob that constructs and sends
a DSS-compliant message to the SQS queue.

* The AWS Ruby SDK requires message attribute keys to be snake case
instead of camel case and symbols instead of strings. It's unclear
if this will cause issues with DSS.
* Added a message to development.rb about the ActiveStorage host.
Currently this needs to be each time one starts a rails console,
but there is an open PR in the rails codebase that will allow us
to configure it for the dev environment.
* We chose to use AWS client stubs for testing purposes rather than
VCR. This means that we'll be instantiating the SQS client outside
of the job, but we don't anticipate any issues with this approach.

#### Developer

#### Code Reviewer

NO

YES

### DIFF
--- a/test/models/dspace_metadata_test.rb
+++ b/test/models/dspace_metadata_test.rb
@@ -10,43 +10,56 @@ class DspaceMetadataTest < ActiveSupport::TestCase
     thesis.save
   end
 
-  test 'parses thesis data as DSpace DC' do
+  test 'metadata file contains thesis data' do
     t = theses(:one)
     dss_friendly_thesis(t)
-    dc = DspaceMetadata.new(t).instance_variable_get(:@dc)
-    assert_equal 'MyString', dc['dc.title']
-    assert_equal '2017-09', dc['dc.date.issued']
-    assert_equal 'MyText', dc['dc.description.abstract']
+    serialized = DspaceMetadata.new(t).serialize_dss_metadata
+    unserialized = JSON.parse(serialized)
+
+    assert unserialized['metadata'].include?({ 'key' => 'dc.title', 'value' => 'MyString' })
+    assert unserialized['metadata'].include?({ 'key' => 'dc.date.issued', 'value' => '2017-09' })
+    assert unserialized['metadata'].include?({ 'key' => 'dc.description.abstract', 'value' => 'MyText' })
 
     # No abstract (optional for undergraduate theses)
     t.abstract = nil
     t.save
-    dc = DspaceMetadata.new(t).instance_variable_get(:@dc)
-    assert_nil dc['dc.description.abstract']
+    dss_friendly_thesis(t)
+    serialized = DspaceMetadata.new(t).serialize_dss_metadata
+    unserialized = JSON.parse(serialized)
+    refute unserialized['metadata'].include?({ 'key' => 'dc.description.abstract', 'value' => '' })
   end
 
-  test 'parses author data as DSpace DC' do
+  test 'metadata file contains author data' do
     # One author
     t = Thesis.create(title: 'Who cares', graduation_year: '2021', graduation_month: 'February',
                       advisors: [advisors(:first)], users: [users(:second)], degrees: [degrees(:one)],
                       departments: [departments(:one)], copyright: copyrights(:mit))
     dss_friendly_thesis(t)
-    dc = DspaceMetadata.new(t).instance_variable_get(:@dc)
-    assert_equal ['Student, Second'], dc['dc.contributor.author']
+    serialized = DspaceMetadata.new(t).serialize_dss_metadata
+    unserialized = JSON.parse(serialized)
+
+    assert unserialized['metadata'].include?({ 'key' => 'dc.contributor.author', 'value' => 'Student, Second' })
 
     # More than one author
     t.users = [users(:second), users(:third)]
     t.save
-    dc = DspaceMetadata.new(t).instance_variable_get(:@dc)
-    assert_equal ['Student, Second', 'Student, Third'], dc['dc.contributor.author']
+
+    dss_friendly_thesis(t)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -73,6 +73,15 @@ of variables:
 In addition, you will need to ensure the bucket CORS `AllowedOrigin` settings
 are configured to allow for the domain this app runs at.
 
+## SQS Configuration
+
+We use AWS SQS queues to publish theses to DSpace and read data about published theses from DSpace. The
+[DSpace Submission Service](https://github.com/MITLibraries/dspace-submission-service) middleware supports this workflow.
+
+`SQS_INPUT_QUEUE_NAME` - The name of the SQS input queue used for publication to DSpace.
```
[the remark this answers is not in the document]
JPrevost: The Python sdk allows us to use queues just by name. I want to see if the ruby sdk does as well as the full URL for the input queue and just the name of the output queue feels weird. It's fine if ruby doesn't have that affordance though.

config/environments/test.rb
```diff
@@ -11,7 +11,9 @@
   ENV['SP_PRIVATE_KEY'] = ''
   ENV['SP_CERTIFICATE'] = ''
   ENV['THESIS_ADMIN_EMAIL'] = 'test@example.com'
-  ENV['OUTPUT_QUEUE_NAME'] = 'test_queue'
+  ENV['SQS_INPUT_QUEUE_URL'] = 'http://localhost:5000/123456789012/test1'
+  ENV['SQS_OUTPUT_QUEUE_NAME'] = 'test2'
```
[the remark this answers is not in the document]
jazairi: We need `SQS_OUTPUT_QUEUE_NAME` to build the message attributes. It's used in the SqsMessage class: https://github.com/MITLibraries/thing/blob/4f2b7bdf9c712f37eab8160c2db3d35a15c7d3ae/app/models/sqs_message.rb#L16

test/models/dspace_metadata_test.rb
```diff
@@ -56,138 +70,251 @@ def dss_friendly_thesis(thesis)
...
+    u.orcid = nil
+    u.save
+    t2.reload
+    serialized = DspaceMetadata.new(t2).serialize_dss_metadata
+    unserialized = JSON.parse(serialized)
+
+    assert unserialized['metadata'].include?({ 'key' => 'dc.identifier.orcid', 'value' => '0001' })
+    refute unserialized['metadata'].include?({ 'key' => 'dc.identifier.orcid', 'value' => '0002' })
 
     # One author and no ORCID
```
matt-bernhardt, line 90: I see what the changes to the test do, and don't object - but this is no longer a test of one author and no ORCIDs - there are two authors in the t2 fixture at this stage, neither of which have an ORCID. The tests seem to confirm that zero ORCIDs are present, which I think is the correct test, but the comment probably needs to be updated.

test/models/dspace_metadata_test.rb
```diff
@@ -56,138 +70,251 @@ def dss_friendly_thesis(thesis)
...
-    assert_equal %w[MFA JD], dc['dc.description.degree']
-    assert_equal ['Master of Fine Arts', 'Master of Fine Arts'], dc['thesis.degree.name']
-    assert_equal %w[Bachelor Master], dc['mit.thesis.degree']
+    dss_friendly_thesis(t)
+    serialized = DspaceMetadata.new(t).serialize_dss_metadata
+    unserialized = JSON.parse(serialized)
+
+    assert unserialized['metadata'].include?({ 'key' => 'dc.description.degree', 'value' => 'MFA' })
+    assert unserialized['metadata'].include?({ 'key' => 'dc.description.degree', 'value' => 'JD' })
+    # TODO: are we supposed to be including the same degree name twice if they are the same?
```
matt-bernhardt, line 274: Is there a ticket in the backlog to resolve this question?

test/models/dspace_metadata_test.rb
```diff
@@ -10,43 +10,57 @@ def dss_friendly_thesis(thesis)
...
     t = theses(:one)
     dss_friendly_thesis(t)
-    dc = DspaceMetadata.new(t).instance_variable_get(:@dc)
-    assert_equal 'MyString', dc['dc.title']
-    assert_equal '2017-09', dc['dc.date.issued']
-    assert_equal 'MyText', dc['dc.description.abstract']
+    serialized = DspaceMetadata.new(t).serialize_dss_metadata
+    unserialized = JSON.parse(serialized)
+
+    assert_equal ['metadata'], unserialized.keys
```
matt-bernhardt, line 19: This test is repeated down in the bottom of the file, and I'm not sure why - I'm not sure it is a huge problem, I've repeated tests as part of separate sequences to confirm the expected, but I'm not sure I follow the distinction between this test and what's on line 330.

matt-bernhardt (commented):
I've looked at only the final commit in this PR, but no earlier work. If needed, I'm happy to look earlier if that would help give context. I'm making this a Comment rather than an Approval only because I want to make sure that I've done the expected scope of work.

The three comments I've made here are non-blocking in themselves. There's a comment in a test that I think needs to be updated, one repeated assertion which I'm not sure is needed, and a TODO comment that I think probably should become a ticket in the backlog.

test/models/dspace_metadata_test.rb
JPrevost: I suspect I copy/pastad the wrong comment. I'll take a closer look. Thanks for catching that.

PR #808, opened 2021-10-29
Enables dspace publication jobs

JPrevost:

* To date, we had all of the pieces complete, but they were not
  connected. This connects all of the existing pieces.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-475

* Creates a new `DspacePublicationPrepJob`
* Configures the existing controller to send the just approved thesis to
  this new job

* A new job was created to loop over the theses that staff have selected
  for publishing and create one job each for the actual publishing. This
  allows the existing publishing job to remain focused on just one thesis
  at a time and since it was already complex enough this seemed better
  than modifying the existing job to do more.
* `DspacePublicationJob` now sets `ActiveStorage::Current.host` in
  development
* `DspacePublicationJob` creates an SQS client if one is not passed in

#### Developer

#### Code Reviewer

NO

NO


app/controllers/thesis_controller.rb
```diff
@@ -56,7 +56,11 @@ def publish_to_dspace
     if params[:graduation] == 'all' || params[:graduation].nil?
       flash[:warning] = 'Please select a term before attempting to publish theses to DSpace@MIT.'
     else
-      flash[:success] = 'If publishing to DSpace@MIT worked, this would be a meaningful update message.'
+      theses = filter_theses_by_term(Thesis.in_review).to_a
```
matt-bernhardt, line 59: This structure - looking up the list in the thesis controller, using the `in_review` scope - makes a lot of sense to me. Could I ask that the commented-out code above this block be removed as part of this PR?

app/jobs/dspace_publication_job.rb
```diff
@@ -1,26 +1,43 @@
...
-    queue_url = ENV.fetch('SQS_INPUT_QUEUE_URL')
+    @queue_url = ENV.fetch('SQS_INPUT_QUEUE_URL')
+    Rails.logger.info("Sending a message to queue #{@queue_url}...")
 
-    Rails.logger.info("Sending a message to queue #{queue_url}...")
+    # We allow the SQS client to be passed in to allow for easier mocking in tests. However, passing the client in via
+    # the controller in dev/prod leads to errors. This is non-ideal, but it allow for us to use a client if one is
+    # passed and create one if not.
+    @sqs_client = sqs_client
+    @sqs_client ||= Aws::SQS::Client.new(region: ENV.fetch('AWS_REGION'))
```
matt-bernhardt, line 14: TIL the `||=` syntax

Would this be less prone to errors if the method tested that the provided client was valid? Or is that just inviting more problems?

app/models/thesis.rb
```diff
@@ -96,7 +96,7 @@ class Thesis < ApplicationRecord
   #  includes(:user).order('users.surname, users.given_name')
   # }
   scope :date_asc, -> { order('grad_date') }
-  scope :in_review, -> { where('publication_status == ?', 'Publication review') }
+  scope :in_review, -> { where('publication_status = ?', 'Publication review') }
```
matt-bernhardt, line 99: D'oh!

config/initializers/sentry.rb
```diff
@@ -0,0 +1,14 @@
+Sentry.init do |config|
+  return unless ENV.has_key?('SENTRY_DSN')
+  config.dsn = ENV.fetch('SENTRY_DSN')
+  config.breadcrumbs_logger = [:active_support_logger, :http_logger]
+
+  # Set tracesSampleRate to 1.0 to capture 100%
+  # of transactions for performance monitoring.
+  # We recommend adjusting this value in production
+  # config.traces_sample_rate = 0.5
```
matt-bernhardt, line 9: I'm assuming that if we need to tune this in production later, we'd do so using ENV vars - but that the volume of this application may allow us to run defaults?

matt-bernhardt (changes requested):
The only comment / request that is blocking is from the thesis controller, asking that commented out code be reviewed - everything else is a question that can be answered whenever.

Generally, this all looks good - but I'd like to not leave commented code when we've already implemented a slightly different approach.

Thanks!

app/jobs/dspace_publication_job.rb
JPrevost: I'll think about checking the client. I'm currently thinking we may not need to check it as I think in reality we'll only pass in a client from the tests and that's a mocked client. In real usage I think we'll always create one here.

app/models/thesis.rb
JPrevost: lol yeah it's funny that sqlite was fine with this whereas Postgres immediately noped out of it

config/initializers/sentry.rb
JPrevost: I don't think we even have access to tracing in our Sentry so this boilerplate comment is probably not super helpful.. but since it was boilerplate I left it.

PR #815, opened 2021-11-08
Adds count of student contributions to dashboard and term report

matt-bernhardt:
This adds information about how many theses have had contributions from students to the reporting dashboard, as well as to the term detail report.

#### Caveats

See the commit message for a list of concerns that I'm having trouble shaking out of my head.

#### Ticket

https://mitlibraries.atlassian.net/browse/ETD-415

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -68,6 +68,14 @@ class Report
     }
   end
 
+  def card_student_contributions(collection)
+    {
+      'value' => collection.map(&:student_contributed?).count(true),
+      'verb' => 'has',
+      'label' => 'had metadata contributed by students'
+    }
+  end
+
   def data_category_copyright_holder
     category = []
     rows = populate_category(Copyright.pluck(:holder))
@@ -211,6 +219,18 @@ class Report
     }
   end
 
+  def data_student_contributions
+    row_data = {}
+    terms = Thesis.all.pluck(:grad_date).uniq.sort
+    terms.each do |term|
+      row_data[term] = Thesis.where('grad_date = ?', term).map(&:student_contributed?).count(true)
+    end
+    {
+      label: 'Students contributing',
+      data: pad_terms(row_data)
+    }
+  end
+
   def data_thesis_records
     {
       label: 'Thesis records',
@@ -243,6 +263,7 @@ class Report
     output['summary'].push data_theses_with_files
     output['summary'].push data_files_attached_to_theses
     output['summary'].push data_issues
+    output['summary'].push data_student_contributions
     output['summary'].push data_multiple_authors
     output['summary'].push data_multiple_degrees
     output['summary'].push data_multiple_departments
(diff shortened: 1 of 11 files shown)


app/models/report.rb
```diff
@@ -211,6 +219,18 @@ def data_multiple_departments
     }
   end
 
+  def data_student_contributions
+    row_data = {}
+    terms = Thesis.all.pluck(:grad_date).uniq.sort
+    terms.each do |term|
+      row_data[term] = Thesis.where('grad_date = ?', term).map(&:student_contributed?).count(true)
```
jazairi, line 226: Just curious, what does passing the boolean to `count` do here? Is it explicitly counting the truthy values resulting from `map(&:student_contributed?`

jazairi (approved):
This looks good! I like the approach of looking at all the versions for student contributions and reporting it in the summary dashboard. My sense is that a detailed view is not required for this, but I'd recommend asking Jess what they need as part of stakeholder review.

I agree that the logic for identifying student submissions/contributions should occur in the Thesis model. I'm not sure I was thinking that through properly in ETD-416. I created a maintenance ticket to move that method: https://mitlibraries.atlassian.net/browse/ETD-521

As for testing, I think you can consider moving the contributor tests from the thesis controller test file to the thesis integration test file. That makes more sense to me conceptually, and it _should_ be some straightforward copypasta, but it's a non-blocking suggestion.

Thanks for this! :shipit:

app/models/report.rb
matt-bernhardt: Yeah, that's a way of counting only the records for which `student_contributed?` is true - without that boolean you get back the list of thesis records for each term. There may be a more efficient way of generating this count, but I'm not sure what it is.

I think there's a ticket in the backlog for dealing with performance of the reporting area, so I'm trying not to worry unduly about performance until that comes up.

matt-bernhardt: Quick addendum - there are three ways that I've handled querying for these reports thus far, but the one in this method is the best one I can come up with at the moment:

1. grab list of terms, iterate over each and select qualifying records (that's the method used here)
2. grab all qualifying records, segmented by term, and then trust in `pad_terms()` to fill in zeros as appropriate. This is the approach taken in `data_thesis_records` below, but I'm unable to make it work for this ticket (yet?)
3. write the underlying SQL ourselves. I try pretty hard to avoid this approach, because it feels fragile and not taking advantage of Rails, but in certain cases it should allow one SQL query to return precisely the records we need. This is seen in the method above this one, `data_multiple_departments`

I'd love to see more methods use option 2, as the code is cleanest in those cases, but I'm not sure that's always the most performant option. I kind of suspect that option 3 is the best for performance, with option 2 as a middle ground and option 1 as the "least performant but most capable" choice.

PR #819, opened 2021-11-12
Update thesis fixtures to have files attached automatically

matt-bernhardt:
This updates our thesis fixtures - as well as one transfer fixture - to have files attached without needing to rely on a `setup` method, or a helper function to do during a test. The work was motivated by some work on ETD-415 and ETD-477, where I realized that having test fixtures with files already attached could be helpful.

Some tests will need files to be attached _during_ the test, so some of these helper functions still remain (and are called).

Additionally, because this work has so far been focused on thesis testing, I've not yet extended this work to update other fixtures - for example, the Registrar fixtures, department tests, and a few others. Searching the test directory for the `.attach` method gives a sense of how much is still being done that way.

I welcome feedback from the reviewer about whether this PR should be extended, and _all_ fixtures updated. At the moment, however, I'm concerned about how much work that would take for something that has not gone through our projects usual ticket triage.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/test/controllers/thesis_controller_test.rb
+++ b/test/controllers/thesis_controller_test.rb
@@ -231,6 +231,13 @@ class ThesisControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
+  test 'processing queue shows nothing without files attached' do
+    sign_in users(:processor)
+    Thesis.all.map { |t| t.files.delete_all }
+    get thesis_select_path
+    assert @response.body.include? 'No theses found'
+  end
+
   test 'processing queue shows records with files attached' do
     sign_in users(:processor)
     get thesis_select_path
--- a/test/controllers/thesis_controller_test.rb
+++ b/test/controllers/thesis_controller_test.rb
@@ -231,19 +231,7 @@ class ThesisControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
-  test 'processing queue shows nothing without work done' do
-    sign_in users(:processor)
-    get thesis_select_path
-    assert @response.body.include? 'No theses found'
-  end
-
-  test 'processing queue shows a record with file attached' do
-    # Attach a file to the thesis
-    t = theses(:with_hold)
-    f = Rails.root.join('test', 'fixtures', 'files', 'a_pdf.pdf')
-    t.files.attach(io: File.open(f), filename: 'a_pdf.pdf')
-
-    # Make sure the url for each thesis processing form is included
+  test 'processing queue shows records with files attached' do
     sign_in users(:processor)
     get thesis_select_path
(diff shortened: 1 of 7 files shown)


test/models/transfer_test.rb
```diff
@@ -15,12 +15,12 @@
 class TransferTest < ActiveSupport::TestCase
   setup do
     @transfer = transfers(:valid)
-    file = Rails.root.join('test', 'fixtures', 'files', 'a_pdf.pdf')
-    @transfer.files.attach(io: File.open(file), filename: 'a_pdf.pdf')
+    # file = Rails.root.join('test', 'fixtures', 'files', 'a_pdf.pdf')
+    # @transfer.files.attach(io: File.open(file), filename: 'a_pdf.pdf')
   end
 
   teardown do
```
jazairi, line 22: I think we can remove this teardown method now that we're not attaching anything in the setup.

test/controllers/thesis_controller_test.rb
```diff
@@ -231,19 +231,7 @@ def attach_files_to_records(tr, th)
     assert_response :success
   end
 
-  test 'processing queue shows nothing without work done' do
```
jazairi, line 234: Even though it's highly unlikely we'll ever see this state, I'm a bit reluctant to remove this test. Maybe we could add something like `Thesis.all.map { |t| t.files.delete_all }` to remove all the file fixtures in the scope of the test?

test/models/transfer_test.rb
```diff
@@ -15,12 +15,12 @@
 class TransferTest < ActiveSupport::TestCase
   setup do
     @transfer = transfers(:valid)
-    file = Rails.root.join('test', 'fixtures', 'files', 'a_pdf.pdf')
-    @transfer.files.attach(io: File.open(file), filename: 'a_pdf.pdf')
+    # file = Rails.root.join('test', 'fixtures', 'files', 'a_pdf.pdf')
+    # @transfer.files.attach(io: File.open(file), filename: 'a_pdf.pdf')
```
jazairi, line 19: Non-blocking, but I'd vote for removing these commented-out lines unless there's a reason we might return to attaching files in the setup..

jazairi (commented):
Excellent work on this! I don't think the ticket should be extended to all fixtures, since that seems like a significant undertaking that probably warrants a ticket. I also don't think any of my comments are blocking. The more I think about it, the less concerned I am about removing that controller test for an empty processing queue.

I'll wait to officially approve until Jeremy has a chance to review, since we decided yesterday that we should all take a look at this. Tentatively, though, it looks :shipit: to me.

matt-bernhardt:
Thanks for your comments - I'll remove the lines from the transfer model test, and look at the suggestion you make for the empty thesis test.

JPrevost (approved):

PR #823, opened 2021-11-19
ETD-502 Fix mailer bug in DSpace Publication Results Job

jazairi:

In its current state, the app does not send mail if the DSpace
Publication Results Job is run and the results include errors only.
This is because mail is sent when results[:processed] is positive,
and this condition will never be met if the job returns no successes.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETD-502

This changes the condition for the DSpace Publication Results Job
to send mail. Mail will now be sent either if results[:total] is
positive, or if results[:errors] has any items. The results[:errors]
piece of this is necessary because the job can fail before it starts
polling, in which case results[:total] will not increment.

I noticed some TODOs in the DSpace Publication Job around setting
the publication status to 'Publication error'. It seems we didn't
have that status in the app when we enabled that job. We do now,
so I updated the job and corresponding tests accordingly.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/jobs/dspace_publication_job.rb
+++ b/app/jobs/dspace_publication_job.rb
@@ -31,7 +31,8 @@ class DspacePublicationJob < ActiveJob::Base
     rescue StandardError, Aws::Errors, Aws::SQS::Errors => e
       Rails.logger.error("Message not sent: #{e}")
       Sentry.capture_exception(e)
-      # Update thesis record with error status once ETD-399 merges
+      thesis.publication_status = 'Publication error'
+      thesis.save
     end
   end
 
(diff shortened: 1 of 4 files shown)


matt-bernhardt (approved):
This all looks :shipit: to me. The change itself makes sense, and I like that we're doing this based on `results[:total]` now, which seems like it gets incremented earlier in the process than `results[:processed]` does (also doing this based on errors is good, obviously.

I do note the CodeClimate warning, but I think I'd prefer to deal with that when we tune logging - the non-logging instructions in that method wouldn't trip the complexity threshold, and it still feels like our logging messages are in the "we are logging lots of things at the start" mode.

:shipit:

PR #826, opened 2021-11-22
ETD-476 Clearly show absence of degrees and advisors in processing form

jazairi:

Theses require degrees and advisors to be valid for publication,
but it's not clear in the UI when a thesis is missing these fields
because the form hides them if they are empty.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETD-476

This adds read-only values for the advisor and degree form elements
that inform the user if the thesis has no associated advisors or
degrees.

* I refactored the degree field UI to a partial, as the additional logic
complicated it enough to warrant it.
* The advisor field refactor removes the association link, since this
creates some complications with the read-only field. I added a hint
with a link to the edit admin path so processors can add advisors
there.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/views/thesis/process_theses.html.erb
+++ b/app/views/thesis/process_theses.html.erb
@@ -160,7 +160,7 @@
 
     <div class="layout-band layout-1q3q field-row">
       <div class="col1q">
-        <strong>Department(s)*</strong>
+        <strong>Department(s) *</strong>
       </div>
       <%= f.simple_fields_for :department_theses do |dept| %>
         <%= render 'department_thesis_fields', :f => dept %>
@@ -173,20 +173,14 @@
     <div class="layout-1q3q layout-band">
       <div class="col3q">Please note: Degree information should be edited via <%= link_to 'the administrative interface', admin_thesis_path(f.object.id), target: :_blank %>.</div>
     </div>
-    <%
-      f.object.degrees.find_all do |deg|
-        if f.object.degrees.count == 1
-          field_id = "degree"
-          field_label = "Degree"
-        else
-          number = (f.object.degrees.find_index(deg).to_i+1).to_s
-          field_id = "degree-" + number
-          field_label = "Degree " + number
-        end
-    %>
+    <% if f.object.degrees.any? %>
+      <% f.object.degrees.find_all do |deg| %>
+        <%= render 'degree_fields', f: f, deg: deg %>
+      <% end %>
+    <% else %>
       <div class="field-row layout-1q3q layout-band">
-        <label class="col1q field-label" for="<%= field_id %>"><%= field_label %></label>
-        <input id="<%= field_id %>" type="text" name="degree[]" readonly="readonly" value="<%= deg.name_dw %>" class="field field-text col3q disabled">
+        <label class="col1q field-label" for="degree">Degree</label>
+        <input id="degree" type="text" name="degree[]" readonly="readonly" value="There are no degrees associated with this thesis." class="field field-text col3q disabled">
       </div>
     <% end %>
 
@@ -211,7 +205,6 @@
         </div>
       </div>
     </div>
-
 
     <%= f.association :copyright, as: :select,
                                   required: true,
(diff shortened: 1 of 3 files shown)


jazairi:
An alternative to the advisor field refactor that I considered is to add a non-form element, styled like a read-only input, that informs the processor that there are no associated advisors. This way, we can safely leave in the link to add new advisors. However, I assume we want to avoid using `link_to_add_association`, since it seems to create duplicate records (e.g., if you add 'Advisorson, Advisor' and there already is an 'Advisorson, Advisor' record, Rails will create a new record rather than finding the duplicate record).

app/views/thesis/process_theses.html.erb
```diff
@@ -160,7 +160,7 @@
 
     <div class="layout-band layout-1q3q field-row">
       <div class="col1q">
-        <strong>Department(s)*</strong>
+        <strong>Department(s) *</strong>
```
matt-bernhardt, line 163: Good catch here

app/views/thesis/process_theses.html.erb
```diff
@@ -240,15 +233,25 @@
...
+        <% end %>
+        <p class="links col3q">
+          <%= link_to_add_association 'Add another supervisor', f, :advisors %>
       </p>
+      <% else %>
+        <div class="advisor nested-fields col3q">
+          <div class="input field_with_hint">
+            <label class="field-label" for="advisors">Supervisor name *</label>
+            <input class="field field-text wide disabled" readonly type="text" value="There are no advisors associated with this thesis." name="advisors" id="advisors">
+            <p class="hint">See details in <%= link_to('the administrative interface.', edit_admin_thesis_path(f.object), target: :_blank) %></p>
```
matt-bernhardt, line 251: I'm not sure that I follow the logic behind not using a `link_to_add_association` link on this branch? The data model was never, IIRC, expecting that advisors would be persistent or a lookup field. The student-facing form just takes arbitrary strings and stores them, and the edit branch of this structure still has one (making it possible to add/drop supervisor strings, including those already defined).

I'm not sure what I'm missing here?

matt-bernhardt (commented):
Almost everything here looks fine - I agree with extracting the degree field out to a partial, and I think I understand why putting the "no degrees" line in a form field is preferable to leaving it in markup (I've confirmed that submitting it with that disabled field value doesn't put anything in that field of the record).

The thing I'm stuck on, which I'd like to hear more about, is why we're not using the "link_to_add_association" on that branch of the conditional, without making any changes to the other forms? It feels inconsistent to ask the processor to change it in the admin area for one condition, but make changes directly on this form in other conditions.

app/views/thesis/process_theses.html.erb
jazairi: The arbitrary string piece of this is what gave me pause. With the form field set to a readonly value, when you add an association, it retains "There are no advisors associated with this thesis" as if it were an actual value. If the processor doesn't delete that value before they save the record, then "There are no advisors associated with this thesis" is stored as an advisor. In this case, the the UI inconsistency seemed less gross than the potential data mess.

A possible alternative here is to make the readonly value a regular div that's styled like a form value. I'm not sure what implications that might have for accessibility, but it's worth a shot. I'll see what that looks like and get back to you.

matt-bernhardt: If it helps, I had to write some javascript for earlier work that hides/shows the add_association links, using some hooks that are available in the gem. I wonder how awful it would be to extend those a little more?

jazairi: Yeah, I suspect the logic would be similar for hiding/showing the placeholder text, so I can probably reuse that. Thanks for pointing that out!

app/views/thesis/process_theses.html.erb
```diff
@@ -240,11 +233,18 @@
...
-      <%= f.simple_fields_for :advisors do |advisor| %>
-        <%= render 'advisor_fields', :f => advisor %>
+      <% if f.object.advisors.any? %>
+        <%= f.simple_fields_for :advisors do |advisor| %>
+          <%= render 'advisor_fields', :f => advisor %>
+        <% end %>
+      <% else %>
+        <div class="nested-fields col3q no-advisors">
+          <label class="field-label" for="advisors">Supervisor name *</label>
+          <input class="field field-text wide input-readonly" readonly type="text" value="There are no advisors associated with this thesis." name="advisors">
```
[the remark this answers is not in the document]
jazairi: I think this is because the input tag was missing an id attribute. This is fixed in the most recent commit, but I can't confirm because accesslint won't run again.

app/views/thesis/process_theses.html.erb
jazairi: @matt-bernhardt I think I have a viable fix for this in place (accesslint struggles notwithstanding). Could you take another look?

app/views/thesis/process_theses.html.erb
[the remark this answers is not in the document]
matt-bernhardt: I'm not seeing an error in either WAVE or ANDI on this field, so I'm inclined to agree that you've fixed the issue.

app/views/thesis/process_theses.html.erb
matt-bernhardt: Yeah, this looks good to me in practice. Thanks!

matt-bernhardt (approved):
This looks good! Thanks for your work on this. :shipit:

PR #828, opened 2021-11-23
Adds color to final banner of publication UI

matt-bernhardt:

* During QA testing for ETD-477, Jess asked that the final banner and
  button that will initiate the publication process be rendered with a
  color, to call out that this is something different than normal
  processing. The goal is to prevent accidental clicking.
* Our current theme gem does not implement a coloring option for this
  "well" element, although elsewhere it does define semantic colors for
  "warning", "error", and "success".

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/etd-477

* This extends our application of the warning, error, and success colors
  to the styles for the "well" elements, so that .well.success will, for
  example, produce a well box that is shaded green. .well.warning will
  render with a yellow treatment, and .well.error will render with a red
  treatment.
* Given these new options, the final publication well is now rendered as
  yellow, or Warning.

* Adding the two color treatments that aren't implemented here - this
  is a hedge against future requests, and the memory that this isn't
  the first time I've wondered about a color option for these elements.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/assets/stylesheets/thing.scss
+++ b/app/assets/stylesheets/thing.scss
@@ -19,7 +19,20 @@ label.error {
   color: $error;
 }
 
-
+.well {
+  &.error {
+    background-color: rgba($error, 0.1);
+    box-shadow: inset 0 0 5px rgba($error, 0.5);
+  }
+  &.success {
+    background-color: rgba($success, 0.1);
+    box-shadow: inset 0 0 5px rgba($success, 0.5);
+  }
+  &.warning {
+    background-color: rgba($warning, 0.1);
+    box-shadow: inset 0 0 5px rgba($warning, 0.5);
+  }
+}
 // admin processing
 .downloaded, .withdrawn {
   @extend .field-wrap;
--- a/app/views/thesis/publish_preview.html.erb
+++ b/app/views/thesis/publish_preview.html.erb
@@ -24,7 +24,7 @@
 </table>
 
 <div id="publish-to-dspace" class="gridband" aria-live="polite" role="region">
-  <div class="inline-action well">
+  <div class="inline-action well warning">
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/thing.scss
```diff
@@ -19,7 +19,20 @@ label.error {
   color: $error;
 }
 
-
+.well {
```
JPrevost, line 22: I'm okay with doing this in-app, but there is argument for this being added upstream to the theme itself

matt-bernhardt: Yeah, I wondered about that too. If you'd like, I can submit this as a PR there.

JPrevost: I'm a bit waffly on this. I'm not against this as-is, but think I prefer it be added upstream (not to the theme gem but to the theme itself as the theme-gem should pull it in automatically on a new release) along with updated examples to use it. If you would prefer to not do that yet, we can land this as-is and open a ticket to move this feature upstream as maintenance in the future (noting that we should remove this when we do it). I'm honestly good with either approach.

matt-bernhardt: Is the theme itself at mitlib-style? I'm realizing that I may not be entirely sure about this setup.

We can continue this discussion on Monday, I think - I may use the intervening time to get a handle on what, exactly, the theme needs to be on-brand. Our panel implementation is different from what I'm envisioning in this work, and I want to consider that a bit more closely.

matt-bernhardt: The more I think about this, the more I kind of want to merge this as-is, and have a ticket in maintenance to add this capability to the theme. I've created https://mitlibraries.atlassian.net/browse/ENGX-105 for that follow-on work, and would like to move forward with merging this at the moment.

PR #833, opened 2021-12-06
Report on expired holds

matt-bernhardt:
This creates a report of holds which should be reviewed for a status change: holds with an end date of today or before, and which have not yet been marked as "released" (i.e. either "active" or "expired"). Each hold displayed is linked to its administrative form, so staff can make any adjustments needed.

We have now gotten some confirmation in the Jira ticket that the assumptions in this work (laid out in the Jira thread) are acceptable to Jess - so I'm marking the "stakeholder approval" box as true.

#### Ticket

https://mitlibraries.atlassian.net/browse/ETD-453

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/controllers/report_controller.rb
+++ b/app/controllers/report_controller.rb
@@ -19,9 +19,7 @@ class ReportController < ApplicationController
   end
 
   def expired_holds
-    report = Report.new
-    holds = Hold.all
-    @list = report.list_expired_holds holds
+    @list = Hold.active_or_expired.ends_today_or_before.order(:date_end)
   end
 
   def files
--- a/app/controllers/report_controller.rb
+++ b/app/controllers/report_controller.rb
@@ -18,6 +18,12 @@ class ReportController < ApplicationController
     @record = report.empty_theses_record subset
   end
 
+  def expired_holds
+    report = Report.new
+    holds = Hold.all
+    @list = report.list_expired_holds holds
+  end
+
   def files
     report = Report.new
     theses = Thesis.all
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -301,10 +301,6 @@ class Report
     }
   end
 
-  def list_expired_holds(collection)
-    collection.active_or_expired.ends_today_or_before.order(:date_end)
-  end
-
   def list_unattached_files(collection)
     result = []
     collection.joins(:files_attachments).order(:grad_date).uniq.each do |record|
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -301,6 +301,10 @@ class Report
     }
   end
 
+  def list_expired_holds(collection)
+    collection.active_or_expired.ends_today_or_before.order(:date_end)
+  end
+
   def list_unattached_files(collection)
     result = []
     collection.joins(:files_attachments).order(:grad_date).uniq.each do |record|
--- a/test/models/hold_test.rb
+++ b/test/models/hold_test.rb
@@ -70,28 +70,47 @@ class HoldTest < ActiveSupport::TestCase
     assert(hold.invalid?)
   end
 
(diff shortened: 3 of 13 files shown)


app/controllers/report_controller.rb
```diff
@@ -18,6 +18,12 @@ def empty_theses
     @record = report.empty_theses_record subset
   end
 
+  def expired_holds
+    report = Report.new
+    holds = Hold.all
```
JPrevost, line 23: Did you consider using the `active_or_expired` scope here directly?

`holds = Hold.active_or_expired`

That would allow for removing that scope from the report itself and I believe end up with the same result.

app/models/report.rb
```diff
@@ -301,6 +301,10 @@ def record_empty_theses(collection)
     }
   end
 
+  def list_expired_holds(collection)
+    collection.active_or_expired.ends_today_or_before.order(:date_end)
```
JPrevost, line 305: Related suggestion to calling the scope in the controller, if we do that I think we could move this to:

`collection.ends_today_or_before.order(:date_end)`

test/models/hold_test.rb
```diff
@@ -70,6 +70,33 @@ class HoldTest < ActiveSupport::TestCase
     assert(hold.invalid?)
   end
 
+  test 'active_or_expired scope shows expected records' do
```
JPrevost, line 73: It might be easier to read these if they were separate tests:

`active_or_expired scope only shows active and expired records`
`active_or_expired scope returns expected active record`
`active_or_expired scope returns expected expired record`
`active_or_expired scope does not return active record`

test/models/hold_test.rb
```diff
@@ -70,6 +70,33 @@ class HoldTest < ActiveSupport::TestCase
...
+    assert Hold.active_or_expired.pluck(:id).include?(hold.id)
+    hold.status = :expired
+    hold.save
+    assert Hold.active_or_expired.pluck(:id).include?(hold.id)
+    hold.status = :released
+    hold.save
+    assert_not Hold.active_or_expired.pluck(:id).include?(hold.id)
+  end
+
+  test 'ends_today_or_before scope shows expected records' do
```
JPrevost, line 87: Similar to above not on tests likely being easier to follow if split up. Alternately adding comments to the tests to show what each assertion is checking would work as well.

JPrevost (commented):
I'm curious your thoughts on whether this needs a report method or not.

The logic itself seems good and is 👍🏻 

Depending on your feelings on refactoring to only use the controller or not we can either merge or make that small change.

Tests comments are just that... comments. Not blocking.

app/controllers/report_controller.rb
matt-bernhardt: Yeah, I had this in an earlier iteration and moved away from it. That is definitely shorter.

My thinking as I undid that work was:
1. Is it really okay to do this work at the controller level? Without instantiating a Report model at all?
2. Other reporting methods in the controller rely on the Report model for their filtering, and I think there is some value in consistency.

These aren't the greatest reasons, I recognize. I'll defer to your judgement on whether a one-line controller is more Rails-y - if so, I'm happy to submit that.

(The consistency line of thought could also result in us doing more at the controller level, particularly for list type reports)

JPrevost: It definitely confused me into thinking I was missing something when I looked and all the report model did for this was use scopes. I think other reports do more complex data formatting in Report which would make the controllers more complex. In this case, using the Reports actually make the controller more complex and the added abstraction seems more like misdirection than useful. However. Neither is correct and this is not my call, it's yours. When there is not a "right" way, it's up to the person submitting to code to decide on which option to use.

JPrevost (approved):

PR #835, opened 2021-12-07
ETD-510 Further describe DSS status errors

jazairi:

When the DSpace Publication Results Job sees a record with a result
type of error, it adds the exception message and the thesis ID to
the errors array. We've seen at least once instance where the result
type is error but the exception message is empty, which results in
an unhelpful message like this: " (thesis ID 42)".

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETD-510

This adds a bit more context to the error message described above,
so can determine the following:

1. That DSS encountered an error while processing the record; and
2. What the publication status was updated to (hopefully 'error').

This will change (i.e., enrich) how errors are reported in the
publication results email.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/jobs/dspace_publication_results_job.rb
+++ b/app/jobs/dspace_publication_results_job.rb
@@ -45,11 +45,12 @@ class DspacePublicationResultsJob < ActiveJob::Base
     when 'success'
       update_handle(thesis, body, results)
     when 'error'
-      error = body['ExceptionMessage']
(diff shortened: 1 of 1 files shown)


JPrevost (approved):

JPrevost (approved):
That error is going to be a mess, but it'll also be the best info we can share. 👍🏻

PR #837, opened 2021-12-09
Warn for Degree and Dept issues when Processing

JPrevost:

* There is no clear indication of issues with associated Departments
  or Degrees on the Thesis processing pages. These issues _will_ prevent
  a Thesis from being published so it makes sense to inform at this
  stage so the Processor can resolve the issues proactively rather than
  find out about them during failed publication attempts.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-517
* https://mitlibraries.atlassian.net/browse/ETD-518

* Creates checks for missing Degree types and Department.name_spaces
* Updates the publication review checks to include these new checks
* Displays messages inline on the Thesis processing page near where
  a processor would be looking at this information anyway so they can
  be aware of problems and resolve them as appropraite.

* We are now preventing the Publication Review status if there are
  blank Department.name_dspace and Degrees missing types
* Fixing a linked record, such as adding a Degree type or Department
  name_dspace does not update the Publication Status. This possibly
  related to this known issue:
  https://mitlibraries.atlassian.net/browse/ETD-507

NOTE: while I did run ANDI and it flagged no concerns around the changes, I do want to consider whether we are adequately highlighting problems with the record in an accessible way. The stakeholders preferred this inline solution, but we may want to reflect on accessibility as a whole on some of these types of interfaces that work by skimming and ensure we are providing the data to assistive devices in a way that also allows skimming. It's possible a more accessible approach would be to include this in the other summary data at the top (or both if stakeholders prefer this solution).

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/thesis.rb
+++ b/app/models/thesis.rb
@@ -140,7 +140,9 @@ class Thesis < ApplicationRecord
       metadata_complete?,
       no_issues_found?,
       no_active_holds?,
-      authors_graduated?
+      authors_graduated?,
+      departments_have_dspace_name?,
+      degrees_have_types?
     ].all?
   end
 
@@ -195,6 +197,20 @@ class Thesis < ApplicationRecord
   def required_license?
     return true if copyright&.holder != 'Author'
     return true if license&.display_description
+
+    false
+  end
+
+  # This returns false if any associated departments are missing dspace names which is a requirment for publishing
+  def departments_have_dspace_name?
+    return true if departments.select { |d| d.name_dspace == '' || d.name_dspace.nil? }.count.zero?
+
+    false
+  end
+
+  # This returns false if any associated degrees are missing types which is a requirement for publishing
+  def degrees_have_types?
+    return true if degrees.select { |d| d.degree_type_id.nil? }.count.zero?
 
     false
   end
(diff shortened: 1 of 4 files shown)


matt-bernhardt (approved):
This all looks good to merge from my perspective, although I wanted to ask about the two approaches taken for providing feedback for degrees and departments. It looks like we are putting a message on the specific degree which has an issue, while for departments the message is more general "one of these has an issue".

Both are fine with me (although I have a slight preference for the degree approach in its clarity), but I'm curious why they are treated differently. This is not a blocking problem, to be clear - I'm just curious.

One other thing I wanted to mention, which is also not blocking, is that we do have publication_status updating for certain types of linked records. Updating a hold should trigger an update to its thesis status changing, via https://github.com/MITLibraries/thing/blob/main/app/models/hold.rb#L76-L80 . The tests for this took me a bit to find, but they are at: https://github.com/MITLibraries/thing/blob/main/test/models/thesis_test.rb#L891-L1001

I don't think this is necessary for this PR, but wanted to mention it when I saw the side effects mentioned in the commit message.

:shipit:

JPrevost:
@matt-bernhardt Thanks for the pointer to how we are updating for Holds. I may try to incorporate that here as well.

As for why Depts and Degrees are different... yeah I agree. In short, the way we construct the form is different for those two. Degrees has left simpleform behind and just created it's own form elements which made this easy to tie in to. Departments is using simpleform natively and getting this to work with that wasn't clear how to do. If this was expected to be a commonly displayed warning, I'd have likely rebuilt the form element to be more like Degrees but seeing as how this should never actually show up unless someone messed up elsewhere... I felt this was close enough.

JPrevost:
@matt-bernhardt upon further reflection, I don't think I want to use the same callback approach we did for Holds here due to the potential difference in scale. A Hold has exactly one Thesis so always saving that one Thesis when we save the Hold is fine. For a Department or Degree change, always saving every associated Thesis is likely hundreds to thousands of saves for one save (per degree period).

I'm going to clean up some of the code climate warnings and leave the updating of all the associated theses to another day.

PR #840, opened 2021-12-09
Add text mapping for "No Creative Commons" license type

jazairi:

During publication workflow testing, stakeholders noted that "No
Creative Commons" looks odd in DSpace, and it would be preferable
for these license types to read as "In Copyright -- Educational Use
Permitted" instead.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETD-514

This adds a method to the license model that returns the desired
phrase for "No Creative Commons" license types, and otherwise
returns the license type as stored in the database. We then use
that method in the Dspace Metadata model to ensure that the mapped
value appears in DSpace.

This change complicates the testing approach we use for the Dspace
Metadata model, which effectively checks for all possible values of
dc.rights in the unserialized metadata against a given thesis'
copyright and license. Now that we are using the same value for two
different copyright/license states, the tests are a little more
ambiguous. This isn't a problem yet, but it's worth documenting here.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/dspace_metadata.rb
+++ b/app/models/dspace_metadata.rb
@@ -1,4 +1,4 @@
-# Generates Dublin Core metadata require to publish a Thesis object to DSpace@MIT via the DSpace Submission System
+# Generates Dublin Core metadata required to publish a Thesis object to DSpace@MIT via the DSpace Submission System
 # DSS. This class assumes that the input Thesis is ready for publication and has all the requisite metadata and file
 # attachments. Validation of those requirements will occur in the Thesis model.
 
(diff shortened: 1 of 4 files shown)


matt-bernhardt (approved):
I see the test suite passing, and the changes made seem reasonable. The testing approach makes sense. It took me a minute to wrap my head around removing the refutations, until I realized that we might as well be also refuting that "The quick brown fox..." was the license statement - or any other arbitrary phrase that has no pathway to be used.

Thanks for this!

PR #841, opened 2021-12-09
Adds a sorting step when attaching files to SQS

matt-bernhardt:

* Bitstreams in DSpace are rendered with the first attachment being
  prioritized, and available for download from the item page. DSS is
  agnostic about the order of files, making ETD responsible for
  attaching them in the correct order (thesis PDF first, then any
  supplementary files)

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/etd-516

* We are already filtering files by their purpose value during the
  creation of the SQS message. This extends that approach, adding a sort
  step to the filtered collection of files. The same array used to
  filter the files is also used (via its index method) to order the
  files appropriately. Only one Thesis PDF is allowed, and the order of
  supplementary files should not matter (famous last words).

* If at some point we have multiple thesis PDFs, or if we need to sort
  the supplementary files, this approach may no longer work.

#### Developer

#### Code Reviewer

NO

NO


matt-bernhardt:
Just a quick note to say that I see the CodeClimate flag, for too many lines in the `map_files` method. Avoiding this is pretty simple, the `@thesis.files.select{...` line can be combined. I'm hesitant to do that, though, and have some lines of this block have multiple commands, but others be separated. I would rather have one command per line, or string them all together (which is too long).

app/models/sqs_message.rb
```diff
@@ -29,7 +29,11 @@ def message_body
   end
 
   def map_files
-    @thesis.files.select { |f| %w[thesis_pdf thesis_supplementary_file].include? f.purpose }.map do |f|
+    ordered_filter = %w[thesis_pdf thesis_supplementary_file]
+    @thesis.files
+           .select { |f| ordered_filter.include? f.purpose }
+           .sort_by { |item| ordered_filter.index(item[:purpose]) }
```
JPrevost, line 35: This works, but I'm slightly worried it's a side effect that makes it work. Our purposes happen to be in the correct alphabetical order so this sorting results in what we currently need. If in the future, we added a new type, such as `thesis_another_not_primary_file_type` that we intend to be publishable this ordering would fail.

I'm not sure yet if we need do an explicit sort to make the `thesis_pdf` the first item in the array or if this semi-side effect is good enough as we could revisit this logic in the future if that theoretical new file type becomes a reality. Thoughts?

jazairi: My understanding of this is that sorting on the `ordered_filter` array's indices ensures that `thesis_pdf` will always come first, so I don't think alphabetization will be a problem. It would be good to confirm this by testing it with a file purpose that would be alpha sorted before `thesis_pdf`, like `proquest_form` or `signature_form`.

matt-bernhardt: I was worried about whether I could trust this method as well, so I did a brief test of this in the console. I _think_ indicates that the sorting isn't alphabetical, but taken from the index of the array:

```
irb(main):068:0> filter_array
=> ["thesis_pdf", "thesis_supplementary_file"]

irb(main):069:0> needle.files.select { |f| filter_array.include? f.purpose }.sort_by { |item| filter_array.index(item[:purpose]) }
=> [#<ActiveStorage::Attachment id: 537, name: "files", record_type: "Thesis", record_id: 24, blob_id: 60, created_at: "2021-08-16 17:43:46.108947000 +0000", description: "", purpose: "thesis_pdf">, #<ActiveStorage::Attachment id: 536, name: "files", record_type: "Thesis", record_id: 24, blob_id: 59, created_at: "2021-08-16 17:43:46.095926000 +0000", description: "", purpose: "thesis_supplementary_file">]

irb(main):070:0> filter_array.reverse
=> ["thesis_supplementary_file", "thesis_pdf"]

irb(main):071:0> needle.files.select { |f| filter_array.include? f.purpose }.sort_by { |item| filter_array.reverse.index(item[:purpose]) }
=> [#<ActiveStorage::Attachment id: 536, name: "files", record_type: "Thesis", record_id: 24, blob_id: 59, created_at: "2021-08-16 17:43:46.095926000 +0000", description: "", purpose: "thesis_supplementary_file">, #<ActiveStorage::Attachment id: 537, name: "files", record_type: "Thesis", record_id: 24, blob_id: 60, created_at: "2021-08-16 17:43:46.108947000 +0000", description: "", purpose: "thesis_pdf">]
```

I'm curious if I'm reading this right - it looks like I get a different sort order if I reverse the order of the filter array.

jazairi: Yeah, that indicates to me that it is in fact sorting on the indexes of the filter array rather than alphabetically. Since we're selecting out any file purposes that aren't in the filter array, I think this approach is solid.

JPrevost: Oh I misread what we were doing! Thanks both of you for clarifying for me. I'd suggest adding that test to save future me from being concerned if you don't mind :)

JPrevost (approved):

PR #844, opened 2021-12-16
Adds a rake task for processing dss output queue

JPrevost:

* We need a way to run the output processing job

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-501

* Creates a rake task to allow us to easily trigger the job to run on
  Heroku from our local CLI rather than having to create a console
  session in Heroku to run job

If multiple apps are connected to the same output queue
(such as local dev and staging), running in one (such as local dev) will
only update the local instance but possibly be processing output for
different environemnts. In practice, this should not be a risk in
production as nobody should have configuration locally to connect to
production output queues so the risk is more associated with PR builds,
Staging and Development environemnts which seem low enough risk to have
the benefits outweigh this concern.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,17 @@ Example usage:
   lines so you'll need to be careful to reconstruct this
 - `rails debug:saml['tmp/your_saml_to_debug.txt']`
 
+## Publishing workflow
+
+- stakeholders process theses until they are valid and accurate
+- stakeholders choose theses to publish (Process theses - Select term - Select Publication Review - Publish)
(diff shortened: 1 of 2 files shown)


jazairi (changes requested):
Thanks for this! It works as expected locally, and I'm not especially concerned about the side effect. Could you add a note to the readme to remind future us that this exists?

jazairi (approved):
Looks good. Thanks! :shipit:

PR #863, opened 2022-01-12
Create structure for BagIt bags

jazairi:

We need a data structure that creates the foundation for preservation
bags.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETD-530

This creates a Bag model that takes a thesis and generates generates
a bag declaration, manifest and bag name for it. It also includes
a `data` method that returns a hash of filepaths for the bag's data
directory, along with their corresponding blobs.

* The base64_to_hex method has been moved to a new Checksums module,
as it is now used both in the publication results job and in this
new model. We may choose to move additional checksum logic to this
module if it is required in multiple places.
* Once ETD-529 is complete, we will need to build the metadata CSV
in this class and add it to the data hash. It's possible this can
all happen within the `data` method, depending on how that ticket
is implemented.
* While this PR meets the need of the ticket conceptually speaking,
additional refinements will likely be required as we determine how
we will create the bags and save them to S3.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/bag.rb
+++ b/app/models/bag.rb
@@ -1,0 +1,44 @@
+# Creates the structure for an individual thesis to be preserved in Archivematica according to the BagIt spec:
+# https://datatracker.ietf.org/doc/html/rfc8493.
+class Bag
+  include Checksums
+
+  def initialize(thesis)
+    @thesis = thesis
+    raise 'This thesis is not baggable' unless baggable?
+  end
+
+  def data
+    file_locations = {}
+    @thesis.files.map { |f| file_locations["data/#{f.filename}"] = f.blob }
+
+    # TODO: build csv, add it to hash
+
+    file_locations
+  end
+
+  def bag_declaration
+    "BagIt-Version: 1.0\nTag-File-Character-Encoding: UTF-8"
+  end
+
+  def manifest
+    @thesis.files.map { |f| "#{base64_to_hex(f.checksum)} data/#{f.filename}" }.join("\n")
+  end
+
+  def bag_name
+    safe_handle = @thesis.dspace_handle.gsub('/', '_')
+    "#{safe_handle}-thesis"
+  end
+
+  # Before we try to bag anything, we need to check if it meets a few conditions. All published theses should have
+  # at least one file attached, no duplicate filenames, and a handle pointing to its DSpace record.
+  # TODO: add a check for metadata csv
+  def baggable?
+    @thesis.files.any? && @thesis.dspace_handle.present? && !duplicate_filenames?
+  end
+
+  def duplicate_filenames?
+    filenames = @thesis.files.map { |f| f.filename.to_s }
+    filenames.select { |f| filenames.count(f) > 1 }.uniq.any?
+  end
+end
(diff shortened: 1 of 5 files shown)


app/models/bag.rb
```diff
@@ -0,0 +1,44 @@
...
+  def bag_name
+    safe_handle = @thesis.dspace_handle.gsub('/', '_')
+    "#{safe_handle}-thesis"
+  end
+
+  # Before we try to bag anything, we need to check if it meets a few conditions. All published theses should have
+  # at least one file attached, no duplicate filenames, and a handle pointing to its DSpace record.
+  # TODO: add a check for metadata csv
+  def baggable?
+    @thesis.files.any? && @thesis.dspace_handle && !duplicate_filenames?
```
JPrevost, line 37: I've noticed an edge case we may need to cover. Locally I had a thesis in which I had manually removed a handle in the UI. The result ends up being `dspace_handle` is an empty string rather than `nil` which this logic considers baggable. I'm not sure how frequently this might come up in production, but I think I'd like to handle it. (ha!). It may be a simple as changing the check to `@thesis.dspace_handle.blank?` which I believe accounts for empty strings, strings with only whitespace, and nil... maybe

jazairi: Good call. Yeah, I think `blank?` should do it. Thanks!

config/application.rb
```diff
@@ -19,5 +19,8 @@ class Application < Rails::Application
     # config.time_zone = "Central Time (US & Canada)"
     # config.eager_load_paths << Rails.root.join("extras")
     config.active_job.queue_adapter = :delayed_job
+
+    # Load modules in /lib
+    config.eager_load_paths += %W(#{config.root}/lib)
```
JPrevost, line 24: I think rather than storing the module in lib and eager loading all of lib (which currently isn't much so it's not a big deal but does include the rake tasks), it may be better to create `app/lib` which is already autoloaded/eagerloaded depending on env and drop our modules in there. Happy to chat about this though.

test/models/bag_test.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  test 'raises exception if the thesis is not baggable' do
+    # baggable thesis (files attached, DSpace handle, no duplicate filenames)
+    thesis = theses(:published)
+    assert_nothing_raised do
+      Bag.new(thesis)
+    end
+
+    # unbaggable thesis (no DSpace handle)
+    thesis.dspace_handle = nil
```
JPrevost, line 77: Let's add a separate check for `thesis.dspace_handle = ""` here as well.

JPrevost (changes requested):
Really good solution to the problem. 🍻 

I added a few comments, one of which is adding a test and handling of an edge case and the other is a suggestion that we might want to move the module to `app/lib` so Rails will handle it's loading for us without us telling it what to do (and possibly being wrong as this stuff is confusing).

I'm a bit waffly on whether we want to throw an exception or just throw to sentry. Let's leave it as is now. The distinction being that if a job runs into one unbaggable thesis, should we stop processing all theses or just that one.

config/application.rb
jazairi: Yeah, that works for me. We could also make it a Concern, but I'm disinclined to do that since we don't need any of the Concern magic.

JPrevost (approved):

PR #864, opened 2022-01-14
Resolve N+1 queries in pub status report and index

JPrevost:

* There were multiple N+1 queries that when combined with production
  data led to Heroku timeouts

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-550
* https://mitlibraries.atlassian.net/browse/ETD-551

* eager load the relevant data to eliminate the N+1 problems

#### Developer

#### Code Reviewer

NO

YES

### DIFF
--- a/app/controllers/thesis_controller.rb
+++ b/app/controllers/thesis_controller.rb
@@ -43,7 +43,7 @@ class ThesisController < ApplicationController
     @terms = defined_terms Thesis.all
     @publication_statuses = Thesis.all.pluck(:publication_status).uniq.sort
     # Filter relevant theses by selected term from querystring
-    term_filtered = filter_theses_by_term Thesis.all
+    term_filtered = filter_theses_by_term Thesis.all.includes(:degrees, :departments, :users)
     @thesis = filter_theses_by_publication_status term_filtered
   end
 
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -223,7 +223,7 @@ class Report
     row_data = {}
     terms = Thesis.all.pluck(:grad_date).uniq.sort
     terms.each do |term|
-      row_data[term] = Thesis.where('grad_date = ?', term).map(&:student_contributed?).count(true)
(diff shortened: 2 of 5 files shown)


app/controllers/thesis_controller.rb
```diff
@@ -43,7 +43,7 @@ def publication_statuses
     @terms = defined_terms Thesis.all
     @publication_statuses = Thesis.all.pluck(:publication_status).uniq.sort
     # Filter relevant theses by selected term from querystring
-    term_filtered = filter_theses_by_term Thesis.all
+    term_filtered = filter_theses_by_term Thesis.all.includes(:degrees).includes(:departments).includes(:users)
```
jazairi, line 46: Is there a difference between this and `.includes(:degrees, departments, users)`?

app/models/report.rb
```diff
@@ -223,7 +223,7 @@ def data_student_contributions
     row_data = {}
     terms = Thesis.all.pluck(:grad_date).uniq.sort
     terms.each do |term|
-      row_data[term] = Thesis.where('grad_date = ?', term).map(&:student_contributed?).count(true)
+      row_data[term] = Thesis.where('grad_date = ?', term).includes(:versions).map(&:student_contributed?).count(true)
```
jazairi, line 226: Is the goal here to reduce the number of queries by loading only theses with papertrail versions?

jazairi (approved):
Neat, TIL `.includes`. This appears to solve the problem, based on some limited local testing. I left a couple of non-blocking questions so I can learn more about this method (happy to discuss in Slack if that's easier).

app/controllers/thesis_controller.rb
JPrevost: No idea! I can try it that way if it reads cleaner to you

app/models/report.rb
JPrevost: Nope. This eager loads the versions so the views don't get into an N+1 situation. Theses without versions should still load I believe (although I'm not sure if that's even a thing)

PR #867, opened 2022-01-18
Initial work on the preservation metadata csv

JPrevost:

* The preservation process will require creating a metadata file
  specific to Archivematica that we will include in the Bag
* Adds the metadata file to the Bag data

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-529

* This work focuses on creating the CSV, but does not yet worry about
  where the file will be created as we aren't sure if we'll save the
  file to S3 (via ActiveStorage or not) or if we'll stream it directly
  into a zip version of the Bag.

#### Developer

#### Code Reviewer

NO

YES


app/models/archivematica_metadata.rb
```diff
@@ -0,0 +1,270 @@
...
+    file_purposes = { 'thesis_pdf' => 'Thesis PDF', 'thesis_source' => 'Thesis source', 'thesis_supplementary_file' =>
+                      'Supplementary file', 'proquest_form' => 'Proquest form', 'signature_page' => 'Signature page' }
+    file_purposes[file.purpose]
+  end
+
+  def bitstream_checksum_value(file)
+    base64_to_hex(file.checksum)
+  end
+
+  # Duplicated logic from dspace_publication_results_job.rb and should be considered form shared location
```
jazairi, line 213: This is now in the Checksums module: https://github.com/MITLibraries/thing/blob/main/app/lib/checksums.rb

app/models/bag.rb
```diff
@@ -12,7 +12,7 @@ def data
     file_locations = {}
     @thesis.files.map { |f| file_locations["data/#{f.filename}"] = f.blob }
 
-    # TODO: build csv, add it to hash
+    file_locations['data/metadata.csv'] = ArchivematicaMetadata.new(@thesis).to_csv
```
jazairi, line 15: This will need a slight rewrite after rebasing, since we're now using an array of arrays in this method instead of a hash.

app/models/bag.rb
```diff
@@ -34,7 +40,7 @@ def bag_name
   # at least one file attached, no duplicate filenames, and a handle pointing to its DSpace record.
   # TODO: add a check for metadata csv
```
jazairi, line 41: I think we can remove this todo, since we're building the csv in the `data` method.

test/models/archivematica_metadata_test.rb
```diff
@@ -0,0 +1,153 @@
+require 'test_helper'
+
+class AuthorTest < ActiveSupport::TestCase
```
jazairi, line 3: I think this should be ArchivematicaMetadatataTest.

test/models/archivematica_metadata_test.rb
```diff
@@ -0,0 +1,153 @@
...
+    thesis.save
+
+    file = Rails.root.join('test', 'fixtures', 'files', 'b_pdf.pdf')
+    thesis.files.attach(io: File.open(file), filename: 'b_pdf.pdf')
+    thesis.files.second.description = 'Not a thesis'
+    thesis.files.second.purpose = 'signature_page'
+    thesis.save
+  end
+
+  # Attaching thesis file so tests will pass
```
jazairi, line 20: Is this comment supposed to be here?

app/models/bag.rb
JPrevost: I don't think that array of arrays work is pushed yet right? So this is a signal to that work to be careful if it moves forward and not for this work to particularly care right?

jazairi: Oh yeah, you're right. My mistake. Too many branches floating around.

PR #869, opened 2022-01-20
Optional settings for test runs

JPrevost:

* we have a few really slow tests that are in isolated pieces of code
  that are useful to have around but are annoying during most
  development

* Adds two environment settings that control the test formating and
  whether to skip tests defined as slow (as of this commit, only one is
  flagged as slow)
* This sets Actions to use a more detailed report of all the tests
  instead of the terse default version and allows developers to use the
  version they prefer locally

* The defaults do not change for either of these new settings, other
  than in Actions where this commit includes explicitly setting the
  test reporter to use the SpecReporter
* The skipped tests are a bit more obvious when the tests run which might be slightly annoying but is probably not a bad thing so we can either remove them or fix them rather than just skip them forever.

#### Developer

#### Code Reviewer

NO

YES

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,8 @@ jobs:
       - name: run tests
         run: |
           bundle exec rails test
+        env:
+          SPEC_REPORTER: true
       - name: Coveralls
         uses: coverallsapp/github-action@v1.0.1
         with:
--- a/Gemfile
+++ b/Gemfile
@@ -53,6 +53,7 @@ end
 
 group :test do
   gem 'climate_control'
+  gem 'minitest-reporters'
   gem 'simplecov', require: false
   gem 'simplecov-lcov', require: false
 end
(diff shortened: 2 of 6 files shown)


jazairi (approved):
Good idea! I'd prefer to fix the test long-term, but this in the meantime this will save a lot of frustration in local dev. Also, I like more verbose descriptions of skipped tests that the `DefaultReporter` provides. It's good to get a reminder of **why** we were skipping those tests, and it makes me want to revisit them. The only downside I can see is the need to change `config/environments/test.rb` locally, and remembering not to check that in to version control, but that's a minor tradeoff. :shipit:

JPrevost:
You don't have to change `test.rb`, you can prefix your test command with a temporary ENV setting like this for Fish shell:

```bash
env SKIP_SLOW=true SPEC_REPORTER=true bin/rails test
```

From my recollection in bash/zsh it's the same but you just leave out the prefix `env`?

jazairi:
Oh yeah, I hadn't considered just changing my shell env. Thanks!

PR #887, opened 2022-02-02
Create and attach zipped bag to SIP

JPrevost:

* delivery of a bag to s3 is how we are handling preservation

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-531
* https://mitlibraries.atlassian.net/browse/ETD-558

* Creates a temporary zip file on disk and streams objects for a thesis
  into that tempfile
* Attaches that tempfile via ActiveStorage to a newly created
  SubmissionInformationPackage

* This uses Heroku's filesystem to store the tempfiles to create the zip
which is non-ideal but because this all happens in a single event it
should be safe to do this. The use of a tempfile ensures the file is
deleted when the handle is closed so if this occurred over multiple
requests it would already fail for non-Heroku reaons.

* This streams objects from S3 into Heroku and back as a zip file. This
will cost data transfer in both directions. A more cost effective
solution if this proves expensive would be to move the logic to create
the zip files to AWS and run it in the same region as the S3 buckets
which would remove the transfer costs.

#### Developer

#### Code Reviewer

NO

YES

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,7 @@ gem 'sentry-ruby'
 gem 'simple_form'
 gem 'skylight'
 gem 'uglifier'
+gem 'zip_tricks'
 
 group :production do
   gem 'pg'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -361,6 +361,7 @@ GEM
     xpath (3.2.0)
       nokogiri (~> 1.8)
     zeitwerk (2.5.4)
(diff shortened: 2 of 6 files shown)


jazairi (changes requested):
Looks good! Sorry I didn't think of this sooner, but I think it makes more sense to pass in a SIP rather than create it on init. Let me know if you have any reservations about that change.

jazairi (approved):

PR #909, opened 2022-02-15
Implement MARC export workflow

jazairi:

We need an automated way to export published records as MARC and
send them to the cataloging team.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETD-564

* Adds a MarcBatch model that builds a compressed marcxml
file from an array of theses.
* Adds a BatchMailer that attaches a compressed marcxml file
built by the MarcBatch model and sends it to the cataloging
team.
* Adds a MarcExportJob that takes an array of theses, builds
a marcxml file for them, and sends a BatchMailer email.
* Updates DspacePublicationResultsJob to run MarcExportJob for
all published theses in the results queue.

* MarcExportJob doesn't employ the 'prep job' pattern that we use
elsewhere. It didn't seem necessary here to differentiate between
processing a single thesis versus an array, since I couldn't
figure a use case for exporting a single thesis. If I'm wrong
about that, this can easily be refactored.
* The tempfiles used to build the compressed marcxml file are
explicitly unlinked in the MarcBatch class, so we pass binary data
to the mailer instead. This felt safer than relying on garbage
collection, but it's a bit less direct than reading the tempfile
in the mailer.

#### Developer

- [x] All new ENV has been added to Heroku Pipeline, Staging and Prod **(with engx email as placeholder value)**

#### Code Reviewer

NO

NO

### DIFF
--- a/app/jobs/marc_export_job.rb
+++ b/app/jobs/marc_export_job.rb
@@ -1,0 +1,20 @@
+class MarcExportJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(theses)
+    marc_filename = "#{filename}.xml"
+    zip_filename = "#{filename}.zip"
+    begin
+      zip_file = MarcBatch.new(theses, marc_filename, zip_filename).build
+      BatchMailer.marc_batch_email(zip_filename, zip_file, theses).deliver_now
+    ensure
+      zip_file.close
+    end
+  end
+
+  private
+
+  def filename
+    "marc_#{DateTime.now.utc.strftime('%y%m%d_%H_%M')}"
+  end
+end
--- a/app/models/marc_batch.rb
+++ b/app/models/marc_batch.rb
@@ -1,0 +1,33 @@
+class MarcBatch
+  def initialize(theses, marc_filename, zip_filename)
+    @theses = theses
+    @marc_filename = marc_filename
+    @zip_filename = zip_filename
+  end
+
+  def build
+    marc_file = Tempfile.new(@marc_filename)
+    zip_file = Tempfile.new(@zip_filename, binmode: true)
+    create_marc_file(marc_file)
+    zip_marc_file(zip_file, marc_file)
+    zip_file
+  end
+
+  private
+
+  def create_marc_file(marc_file)
+    writer = MARC::XMLWriter.new(marc_file.path)
+    @theses.each do |thesis|
+      record = Marc.new(thesis)
+      writer.write(record.record)
+    end
+    writer.close
+  end
+
+  def zip_marc_file(zip_file, marc_file)
+    Zip::File.open(zip_file.path, Zip::File::CREATE) do |zip|
+      zip.add(@marc_filename, marc_file.path)
+    end
+    marc_file.close
+  end
+end
--- a/test/jobs/marc_export_job_test.rb
+++ b/test/jobs/marc_export_job_test.rb
@@ -1,0 +1,24 @@
+require 'test_helper'
+
+class MarcExportJobTest < ActiveJob::TestCase
+  include ActionMailer::TestHelper
+
+  test 'sends batch email' do
+    ClimateControl.modify DISABLE_ALL_EMAIL: 'false' do
+      theses = [theses(:one)]
+      assert_emails 1 do
+        MarcExportJob.perform_now(theses)
+      end
+    end
+  end
+
+  test 'sent email attachments use expected filename format' do
+    ClimateControl.modify DISABLE_ALL_EMAIL: 'false' do
+      theses = [theses(:one)]
+      Timecop.freeze(Time.utc(2022, 2, 14, 17, 10, 0)) do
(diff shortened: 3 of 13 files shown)


test/mailers/batch_mailer_test.rb
```diff
@@ -0,0 +1,22 @@
...
+      # Send the email, then test that it got queued
+      assert_emails 1 do
+        email.deliver_now
+      end
+
+      # Make sure it was sent to the right person with the expected attachment.
+      assert_equal ['test@example.com'], email.from
+      assert_equal ['test-metadata@example.com'], email.to
+      assert_equal 'ETD MARC batch export', email.subject
+      assert_equal 'marc.zip', email.attachments.first.filename
```
JPrevost, line 18: I think this tests the attachment filename effectively enough to remove the test from `MarcExportJobTest` that you had a comment on being kinda weird maybe?

JPrevost (approved):

test/mailers/batch_mailer_test.rb
jazairi: It's a bit different since we're not actually calling `MarcExportJob#filename` here.

app/models/marc_batch.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def zip_marc_file(zip_file, marc_file)
+    Zip::File.open(zip_file.path, Zip::File::CREATE) do |zip|
+      zip.add(@marc_filename, marc_file.path)
+    end
+    zip_file.close
+    marc_file.unlink
+
+    # We are reading the file here so we can unlink the tempfile and pass the binary data directly to the mailer.
```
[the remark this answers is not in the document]
jazairi: Yeah, that's a good point. I'll take a closer look at how ruby handles tempfiles before we move forward with this approach.

test/jobs/marc_export_job_test.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+  end
+
+  test 'enqueued email attachments use expected filename format' do
+    theses = [theses(:one)]
+    Timecop.freeze(Time.utc(2022, 2, 14, 17, 10, 0)) do
+      email = MarcExportJob.perform_now(theses)
+      
+      # A bit byzantine...is there a better way to do this?
+      fname = email.instance_variable_get(:@arguments).last[:args].first
```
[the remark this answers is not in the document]
jazairi: Same. I suppose I'll leave it in (without the quizzical comment) in the interest of more coverage.

app/models/marc_batch.rb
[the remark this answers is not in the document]
jazairi: @JPrevost It occurred to me that we can just close/unlink the tempfile in an `ensure block` in the job. However, ActiveJob doesn't accept tempfiles as arguments, meaning we wouldn't be able to use `deliver_later` to send the email. We could use `deliver_now`, since that sends the email immediately without enqueueing it. Apparently we already use `deliver_now` to send the publication results email, though I'm not sure why: https://github.com/MITLibraries/thing/blob/main/app/jobs/dspace_publication_results_job.rb#L23

Thoughts on this? It seems to me that any risk in using `deliver_now` is outweighed by the benefits of not reading the zip file into memory.

jazairi:
@JPrevost I just pushed a refactor to use the tempfiles as attachments. Could you re-review?

app/jobs/marc_export_job.rb
```diff
@@ -4,8 +4,8 @@ class MarcExportJob < ActiveJob::Base
   def perform(theses)
     marc_filename = "#{filename}.xml"
     zip_filename = "#{filename}.zip"
-    zip_data = MarcBatch.new(theses, marc_filename, zip_filename).build
-    BatchMailer.marc_batch_email(zip_filename, zip_data, theses).deliver_later
+    zip_file = MarcBatch.new(theses, marc_filename, zip_filename).build
+    BatchMailer.marc_batch_email(zip_filename, zip_file, theses).deliver_now
```
JPrevost, line 8: I wonder if doing something like:
```ruby
begin
  zip_file = MarcBatch.new(theses, marc_filename, zip_filename).build
  BatchMailer.marc_batch_email(zip_filename, zip_file, theses).deliver_now
ensure
  zip_file.close
end
```

...rather than closing from the BatchMailer might both work and be easier to keep track of. Note: I have not tried this and there may be reasons it's not the way to go :)

jazairi: @JPrevost Yeah, I do like that better. My only hesitation is I think that would make the explicit close untestable, since `zip_file` isn't returned anywhere. Do you have any concerns about that?

JPrevost: Not really. I feel like we can prove to ourselves this works by looking at our filesystem to ensure the file gets removed when we expect and then trust ruby to be consistent. Thanks for noting that concern though.

JPrevost (approved):

PR #937, opened 2022-03-17
Fix publication status update logic

jazairi:

When a thesis is saved, Thesis#evaluate_status will not evaluate as
expected, due to certain convenience methods used within
`Thesis#required_fields?`. These methods return the wrong value
before the record is saved, requiring processors to submit the
processing form twice in order to update a thesis' publication status.

Because we use the `before_save` callback, this bug can also occur if
nested attributes are updated, but the Thesis model is not (e.g.,
a new degree is attached, but none of the thesis attributes are
changed). This means that the thesis record is not technically saved,
and the `before_save` callback won't fire.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETD-507

* Updates the logic in `Thesis#advisors?` and `Thesis#degrees?`
to use `any?` rather than `count.zero?`, which evaluates correctly
before the thesis is saved.
* Uses the `after_validation` ActiveRecord callback instead of
`before_save`, so `Thesis#update_status` will be called when only
nested attributes are updated.

* Refactored `Thesis#files?` to mirror the logic in `#advisors?`
and `#degrees?`.
* Removed the `valid?` call from `Thesis#evaluate_status` to avoid
an infinite loop. This call seemed redundant since `validate` is
called before `save`, but I'd be interested in discussing this
in case I'm missing an edge case.
* This feature uses many other convenience methods, some of which
could be refactored or possibly removed. I opened a ticket to
address this here: https://mitlibraries.atlassian.net/browse/ETD-578

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/thesis.rb
+++ b/app/models/thesis.rb
@@ -92,7 +92,8 @@ class Thesis < ApplicationRecord
 
   VALID_MONTHS = %w[February May June September].freeze
 
-  before_save :combine_graduation_date, :update_status
+  before_save :combine_graduation_date
+  after_validation :update_status
   after_find :split_graduation_date
 
   scope :date_asc, -> { order('grad_date') }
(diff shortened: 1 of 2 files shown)


JPrevost (approved):
🍻

PR #1032, opened 2022-08-26
Refactor administrate authorization to support gem update

jazairi:

The most recent version of administrate refactors authorization
in a way that has downstream effects. See [#1941](https://github.com/thoughtbot/administrate/pull/1941)
for more info.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ENGX-192

This replaces our override of the `show_action?` method with
`authorized_action?` in the administrate application controller,
and it replaces our combined check of `valid_action?` and
`show_action?` with `accessible_action?` in the administrate views.

This also updates the ability so thesis processors can continue to
access the submitter dashboard. This is not currently needed
as we do not have any users with the thesis_processor role, but it
seemed wise to address it sooner than later.

* Thesis processors can now see links to controller actions in the
submitter admin dashboard, as well as a link to the admin dashboard
itself from the main site nav.
* The admin submitter dashboard test will move to the admin/
subdirectory with the other individual dashboard tests. This change
will come in a subsequent commit to make the git history easier
to read.

#### Developer

#### Code Reviewer

NO

YES


JPrevost (approved):
I confirmed a few roles manually to confirm they still work as expected which along with our automated tests I feel confident this change is solid.

The weirdness in Submitter ability seems like a bug that you are fixing here so I'm not as worried about that as I was before looking at the change and poking at it myself.

app/controllers/admin/application_controller.rb
```diff
@@ -10,21 +10,23 @@ class ApplicationController < Administrate::ApplicationController
...
-      return if can?(action_name, resource_name)
+      return if authorized_action?(resource_name, action_name)
 
       redirect_to root_path, alert: 'Not authorized.'
     end
 
     # Hide links to actions if the user is not allowed to do them.
     # This is an override of an Administrate method to work with CanCan
-    def show_action?(action, resource)
+    def authorized_action?(resource, action)
```
matt-bernhardt, line 29: Non-blocking: I see here, and in the underlying PR against the gem, that some orders of arguments are changing. This is a bit confusing to me, as the `can?` call is still in the original order, but now some methods go the other way.

I don't think there's anything to be done about this here - I see that `authorized_action?` here, and `accessible_action?` in the views are now consistent in their argument use, which probably hides the complexity well enough - and I don't want to bike-shed this. I'm just sure that this will trip someone up in the future, and I don't look forward to that being me.

matt-bernhardt (approved):
I see the impact of this change when I run the app locally, and think this is good - it should open us up to actually using the processor role in the future, and it also allows us to keep current with updates. There's an aspect I noted in one area that will likely trip someone up, but I don't see a way to resolve this entirely - there seems to be an inconsistency in how arguments are ordered when passed to the gem now.

:shipit:

app/controllers/admin/application_controller.rb
jazairi: Yeah, I hear you. It seemed less confusing to me to match the argument order with what's in the upstream code, but you're absolutely right that this is likely to confuse future us anyway. Fwiw, I tend to think that will unfortunately be the case with any administrate update that involves changing custom controller methods and views.

PR #1040, opened 2022-09-23
Initial set of reporting features for ProQuest opt-in status

jazairi:

Processors need to be able to review ProQuest opt-in status
information at two points in the processing workflow:

1. Prior to publication but after files have been attached,, so they
can review and resolve discrepancies in opt-in status for theses with
multiple authors.
2. After publication, so they can review summary data by
department, degree, and term.

During grooming, these use cases were conceived as separate reports,
but combining them into a single report seemed to be a more
usable design.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-591
* https://mitlibraries.atlassian.net/browse/ETD-599

This adds a report that lists the ProQuest opt-in status for theses
with files attached. The report includes five filters:

* Graduation date (i.e., term)
* Department
* Degree
* Multiple authors (boolean)
* Published theses (boolean)

Additionally, the report displays cards listing summary data for each
status. If a status is included in the filtered results, then a card
will list the number of theses with that status that match all of
the conditions.

* As this is a slight departure from what we discussed in grooming,
it may not be the optimal approach for stakeholders and may need
to be redesigned pending QA.
* The summary data cards do not hook into the existing card system
in the Report model. This particular use case was custom enough that
it seemed more efficient to start from scratch.
* Some existing fixtures have been altered, requiring unrelated
count-based tests to be modified.
* Assertions were added to some unrelated report controller tests
that didn't seem to test anything.

#### Developer

#### Code Reviewer

NO

NO


jazairi:
Note that this is branched off of etd-587-proquest-opt-in. We are waiting to merge that until stakeholders confirm that this feature set is ready to launch.

test/helpers/thesis_helper_test.rb
```diff
@@ -76,4 +76,150 @@ class ThesisHelperTest < ActionView::TestCase
     hybrid_thesis = Thesis.new(degrees: [degrees(:one), degrees(:two)])
     assert_equal true, satisfies_advanced_degree?(hybrid_thesis)
   end
+
+  test 'filter_theses_by_department returns a filtered set of theses' do
+    theses = Thesis.includes(:departments)
+    params[:department] = 'Department of Aeronautics and Astronautics'
+    assert_not_equal theses.count, filter_theses_by_department(theses).count
```
JPrevost, line 83: This assertion isn't wrong, but I wonder if asserting the filtered set is smaller than the full set would provide additional value?

test/helpers/thesis_helper_test.rb
```diff
@@ -76,4 +76,150 @@ class ThesisHelperTest < ActionView::TestCase
...
+  test 'filter_theses_by_degree does nothing when degree param is "all"' do
+    theses = Thesis.includes(:degrees)
+    params[:degree] = 'all'
+    assert_equal theses.count, filter_theses_by_degree(theses).count
+  end
+
+  test 'filter_theses_by_multiple_authors returns a filtered set of theses' do
+    theses = Thesis.all
+    params[:multi_author] = 'true'
+    assert_not_equal theses.count, filter_theses_by_multiple_authors(theses).length
```
JPrevost, line 117: Similar to another comment, consider whether checking for a smaller set rather than not equal would be more meaningful

test/helpers/thesis_helper_test.rb
```diff
@@ -76,4 +76,150 @@ class ThesisHelperTest < ActionView::TestCase
...
+
+  test 'filter_theses_by_multiple_authors does nothing when no multi_author param is set' do
+    theses = Thesis.all
+    assert_equal theses.count, filter_theses_by_multiple_authors(theses).length
+  end
+
+  test 'filter_theses_by_published returns a filtered set of theses' do
+    theses = Thesis.all
+    params[:published] = 'true'
+    assert_not_equal theses.count, filter_theses_by_published(theses).count
```
JPrevost, line 134: Similar to another comment, consider whether checking for a smaller set rather than not equal would be more meaningful

JPrevost (commented):
I think this might need a bit more work, or the data in the PR build is just weird (equally likely lol):

https://thesis-submit-pr-1040.herokuapp.com/report/proquest_status

I can see a report that shows 3 users have opted in (I adjusted authors manually to reflect that).

But choosing published yes or published no always results in zero theses showing.

I'll hold off on further reviewing functionality until you get a chance to look into that. Thanks!

test/controllers/report_controller_test.rb
```diff
@@ -772,4 +778,179 @@ def create_thesis_with_whodunnit(user)
...
+    assert_select 'table tbody tr', count: thesis_count
+  end
+
+  test 'ProQuest status report allows filtering by term' do
+    thesis_count = Thesis.with_files.count
+    term_count = Thesis.with_files.where('theses.grad_date = ?', '2021-06-01').count
+    assert_not_equal thesis_count, term_count
+    assert term_count > 0
+
+    # Add 1 to the select counts for filter options
```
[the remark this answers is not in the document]
jazairi: The default option is 'all', so it's the actual filter options plus the default.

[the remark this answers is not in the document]
JPrevost: I'd still prefer we update these comments to explain _why_ we're adding 1, not _that_ we're adding 1 which we know by just reading the code.

Changing the comment to something like "Adding 1 to reflect all options in the filter plus the default ALL option" or similar would require less guessing from a reader of this test. As is, the comment just says what the code does, not why, which wasn't helpful to me.

db/schema.rb
```diff
@@ -10,13 +10,16 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_08_26_202539) do
+ActiveRecord::Schema.define(version: 2022_10_17_163325) do
+
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "plpgsql"
```
JPrevost, line 16: @jazairi can you revert this file locally and rerun your migrations with your dev env set to sqlite instead of pg? I'm curious if some of these changes are introduced due to your temp setup with this work requiring staging data.

jazairi: Yeah, that appears to be what happened. Thanks for the catch. Just pushed up a commit addressing that.

app/views/report/_proquest_thesis.html.erb
```diff
@@ -0,0 +1,10 @@
+<tr>
+  <td><%= link_to title_helper(proquest_thesis), edit_admin_thesis_path(proquest_thesis.id) %></td>
+  <td>
+    <% proquest_thesis.authors.each do |author| %>
+      <%= link_to author.user.display_name, edit_admin_author_path(author.id) %><br>
```
JPrevost, line 5: This is super helpful. I've noticed some of our views don't link usefully to where you can view/edit the data driving the view.

The resulting administrate edit feels a bit weird though as it only has a checkbox which doesn't seem to fully account for the various statuses that we imply are available:

- yes, the user definitely opted-in
- no, the user definitely opted-out
- the user didn't edit the data at all so it's null

I think if a staff member goes to that form and then just clicks "save" the result will change from "null" no "no" implying a user _did_ make a choice which isn't accurate (whether that matters is unclear). Thoughts on whether we can have administrate more accurate reflect the yes/no/null options on that form?

db/migrate/20221017163325_reset_theses_cache_counters.rb
```diff
@@ -0,0 +1,10 @@
+class ResetThesesCacheCounters < ActiveRecord::Migration[6.1]
+  def up
+    Thesis.all.each do |thesis|
+      Thesis.reset_counters(thesis.id, :authors)
```
JPrevost, line 4: I'm semi-nervous to allow this to happen during the pre-deploy step. It's not wrong by any means and will do what we need, but I sort of prefer to let the app start normally and then console in to do this one-time thing with the slight downside that the reports would have wrong data for the few minutes that it takes us to console in and run the script. This isn't a blocking comment, just expressing my "hmmm" thought on the two approaches.

test/fixtures/authors.yml
```diff
@@ -24,11 +25,13 @@ two:
   user: yo
   thesis: two
   graduation_confirmed: true
+  proquest_allowed: true
 
 three:
   user: basic
   thesis: two
   graduation_confirmed: false
+  proquest_allowed: false
```
JPrevost, line 34: Are we also accounting for the `null` value in any tests?

test/models/report_test.rb
```diff
@@ -48,7 +48,7 @@ class ReportTest < ActiveSupport::TestCase
     result = r.index_data
     assert_equal Department.count, result['departments'].pluck(:label).length
     assert_includes result['departments'].pluck(:label), Department.first.name_dw
-    assert_equal result['departments'][0][:data].values, [0, 3, 0, 0, 0, 0, 0, 4, 0]
+    assert_equal result['departments'][0][:data].values, [0, 2, 0, 0, 0, 0, 0, 3, 1]
```
JPrevost, line 51: This values changed and I have no idea what they mean and if they should have. I'm okay with not understanding these tests for now if you have thought about why these values changed and they are appropriate (this applies to all of the changing numbers in this test file).

app/views/report/_proquest_thesis.html.erb
jazairi: Yes, we should address this; null and false are very different values in this case. One potential solution could be to change the field to to a select, so that null would be one of the options in the drop-down menu.

test/models/report_test.rb
jazairi: I stared at this for about an hour, and I'm 75% sure that these changes are due to the changes in the fixtures.

test/fixtures/authors.yml
jazairi: We are in the thesis_helper_test coverage for the ProQuest status methods: https://github.com/MITLibraries/thing/blob/6050cf75e7fcce10b7dda28432726f340d25a522/test/helpers/thesis_helper_test.rb#L193-L224

jazairi: Oh neat, I didn't realize that GitHub now translates urls to code snippets!

db/migrate/20221017163325_reset_theses_cache_counters.rb
jazairi: As discussed in our team meeting, I'm planning to replace this with a rake task. In the unlikely event that we have to reset counters more than once, having a documented rake task will help us remember how (and why) to do it.

JPrevost: The solution you put in works great. Thanks!

JPrevost: Ha, I meant that for the administrate bit... but the rake task is great too :)

JPrevost (approved):
Thanks for pushing through this, I know we way underestimated the effort it would be.

👍🏻

PR #1097, opened 2022-12-23
Add batch export feature for ProQuest JSON

jazairi:
Why these changes are being introduced:

Processors and thesis admins need to be able to export a list of thesis handles that have opted in to ProQuest. An additional export is needed for doctoral theses that have opted out of ProQuest (or have not responded). No additional filters are required for these exports.

Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETD-588
https://mitlibraries.atlassian.net/browse/ETD-603

How this addresses that need:

This adds a view that allows processors to review theses that are ready to be exported to ProQuest. Two types of theses meet this criterion:

* Published theses satisfying an advanced degree, with all authors having consented to ProQuest export. ProQuest will fully harvest these theses (metadata and content).
* Published theses satisfying a doctoral degree, without all authors having consented to ProQuest export. ProQuest wil partially harvest these theses (metadata only).

Once a processor has reviewwd the list of theses to be exported, they can initiate the export. This will kick off a background job that will produce a JSON file that lists each thesis' handle and distinguishes between full and partial exports. Once the export is ready, the job send an email to the ETD admin list with the JSON attached.

Finally, all exported theses will be associated with a ProquestExportBatch, and the JSON file will be attached to the same as an ActiveStorage attachment.

Side effects of this change:

* ProquestExportJob currently loops through theses to update them using the
`each` method. This is an expensive operation. A more efficient option is
update_all, which updates attributes without instantiating the objects.
However, this would not trigger callbacks, so we would need to find another
way to get paper_trail to update, such as this hack:
https://github.com/paper-trail-gem/paper_trail/issues/456#issuecomment-1085583174
* This adds a has_many/belongs_to relationship between the ProquestExportBatch
and Thesis models. This may not be necessary since we are storing the JSON
exports in ActiveStorage, but it could be useful to look up the date a thesis
was exported and find other theses from the same export.
* The proquest_exported field has been added to the thesis admin dashboard.
* The _proquest_status_empty partial is moved to views/shared as it is now
used by the export preview and the export report.
* Some unrelated tests in the Report model is updated due to the addition
of a new authors fixture.

#### Developer

#### Code Reviewer

YES

NO

[ETD-588]: https://mitlibraries.atlassian.net/browse/ETD-588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ETD-603]: https://mitlibraries.atlassian.net/browse/ETD-603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/views/thesis/proquest_export_preview.html.erb
```diff
@@ -0,0 +1,56 @@
...
+<%= render 'shared/you_are' %>
+
+<div class="gridband" id="export-action" aria-live="polite" role="region">
+  <div class="inline-action well warning">
+    <% if @theses.count < 1 %>
+      <p>No theses are available to export.</p>
+    <% else %>
+      <div class="message">
+        <h4 class="title">Ready to export <%= @theses.count %> theses?</h4>
+        <p>If the set of thesis records below looks correct, click the button at right to export their handles. Use the
```
JPrevost, line 20: Small wording suggestion: `If the set of thesis records below is correct...`. I think it may be worthwhile to be clearer that the accountability of the record set is on the person clicking `export`

test/controllers/thesis_controller_test.rb
```diff
@@ -769,4 +769,122 @@ def attach_files_to_records(tr, th)
...
+    get thesis_proquest_export_path
+    assert_enqueued_jobs 1
+    assert_redirected_to thesis_proquest_export_preview_path
+    follow_redirect!
+    assert_select '.alert-banner.success', text: 'The theses you selected will be exported. Status updates are not immediate.', count: 1
+  end
+
+  test 'exporting to proquest with no eligible theses redirects to proquest export preview path with a flash message' do
+    sign_in users(:processor)
+    Thesis.destroy_all
```
JPrevost, line 884: See previous note about `delete_all`

test/models/proquest_export_batch_test.rb
```diff
@@ -0,0 +1,7 @@
+require "test_helper"
+
+class ProquestExportBatchTest < ActiveSupport::TestCase
+  # test "the truth" do
```
JPrevost, line 4: It might be useful to confirm the json builds correctly here? Not sure.

test/models/thesis_test.rb
```diff
@@ -1245,4 +1245,100 @@ def attach_file_with_purpose_to(thesis, purpose = 'thesis_pdf')
...
+    # multi-author thesis that has opted out is excluded
+    thesis = theses(:with_hold)
+    assert_not_includes Thesis.consented_to_proquest, thesis
+
+    # multi-author thesis with conflicting opt-in statuses is excluded
+    thesis = theses(:doctor)
+    assert_not_includes Thesis.consented_to_proquest, thesis
+  end
+
+  test 'invalid without proquest_exported' do
```
JPrevost, line 1278: This test name may benefit from some clarity. None of the assertions seem to result in `invalid` which I'd expect based on this test name.

test/models/thesis_test.rb
```diff
@@ -1245,4 +1245,100 @@ def attach_file_with_purpose_to(thesis, purpose = 'thesis_pdf')
...
+    thesis.proquest_exported = 'Full harvest'
+    thesis.save
+    assert thesis.valid?
+
+    thesis.proquest_exported = 'Partial harvest'
+    thesis.save
+    assert thesis.valid?
+  end
+
+  test 'not_consented_to_proquest returns only theses with authors that have not opted in to ProQuest' do
```
JPrevost, line 1292: Thanks for the clear comments as to all of the conditions being tested here!

test/models/thesis_test.rb
```diff
@@ -1245,4 +1245,100 @@ def attach_file_with_purpose_to(thesis, purpose = 'thesis_pdf')
...
+    partially_harvested_thesis = theses(:doctor)
+    assert_includes Thesis.exported_to_proquest, partially_harvested_thesis
+  end
+
+  test 'exported_to_proquest includes fully harvested theses' do
+    fully_harvested_thesis = theses(:engineer)
+    assert_includes Thesis.exported_to_proquest, fully_harvested_thesis
+  end
+
+  test 'exported_to_proquest excludes theses that have not been harvested' do
```
JPrevost, line 1328: The wording on this (and possibly the two previous tests but I haven't decided) as well as the next few may be awkward. We are only in control of when we export theses to Proquest not when they actually get around to Harvesting them. The enum values themselves are fine as we export them as "Full harvest"able.

My concern is that in the future these tests names may imply that we somehow are updating/knowing when Proquest comes back and harvests theses which is not the case.

Rewording to something like `exported_to_proquest excludes theses that are flagged not harvestable` (I'm not sure if that is entirely correct wording but I hope it provides a useful example of what I'm trying to convey)

test/controllers/thesis_controller_test.rb
```diff
@@ -769,4 +769,122 @@ def attach_files_to_records(tr, th)
...
+
+  test 'admins can load the proquest export preview list' do
+    sign_in users(:admin)
+    get thesis_proquest_export_preview_path
+    assert_response :success
+  end
+
+  test 'proquest export cannot be initiated with no eligible theses' do
+    sign_in users(:processor)
+    Thesis.destroy_all
```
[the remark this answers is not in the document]
jazairi: Oh, good call.

app/controllers/thesis_controller.rb
```diff
@@ -134,6 +134,30 @@ def process_theses_update
     redirect_to thesis_process_path
   end
 
+  def proquest_export_preview
+    advanced_theses = Thesis.where(publication_status: 'Published').advanced_degree.not_exported_to_proquest
```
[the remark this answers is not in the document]
jazairi: Yeah, this felt gross even as I wrote it. The job expects three parameters so it can separately update the sets of partial and full harvest thesis records. I don't think a scope would work because the logic is too complex, but a class method is a good idea.

app/jobs/proquest_export_job.rb
```diff
@@ -0,0 +1,31 @@
+class ProquestExportJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(all_to_export, full_harvest, partial_harvest)
+    export = ProquestExportBatch.new
+
+    # update_all is an option here if performance is poor, but we would need a workaround to update paper_trail
+    # versions as update_all does not trigger callbacks.
+    if partial_harvest.present?
```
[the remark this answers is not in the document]
jazairi: Alternatively, I think `Thesis.partial_export` and `Thesis.full_export` scopes would achieve the same.

app/views/thesis/_proquest_export_thesis.html.erb
```diff
@@ -0,0 +1,11 @@
+<tr>
```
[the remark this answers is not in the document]
jazairi: The fields displayed. This one has `grad_date` instead of the ProQuest opt-in status.

test/models/thesis_test.rb
jazairi: Agreed. I'll clarify the language in these.

test/models/proquest_export_batch_test.rb
jazairi: Just as I try to introduce a little danger into our test suite... (I totally forgot to write any tests for this model; thanks for catching!)

app/views/thesis/_proquest_export_thesis.html.erb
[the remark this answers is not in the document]
JPrevost: lol doh. Thanks!

app/jobs/proquest_export_job.rb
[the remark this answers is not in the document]
JPrevost: And then just call them directly instead of passing them in? Yeah that could work. That may introduce a very unlikely race condition I suppose...

test/controllers/thesis_controller_test.rb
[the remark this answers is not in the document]
jazairi: Ah, that's right. We can't use `delete_all` because we need  the `dependent: :destroy` callbacks, or else these tests will throw foreign key constraint exceptions.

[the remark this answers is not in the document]
JPrevost: 🙈

test/models/proquest_export_batch_test.rb
```diff
@@ -1,7 +1,15 @@
 require "test_helper"
 
 class ProquestExportBatchTest < ActiveSupport::TestCase
-  # test "the truth" do
-  #   assert true
-  # end
+  test 'builds JSON with expected values' do
```
JPrevost, line 4: Thanks for adding these tests!

JPrevost (commented):
I've finished the code review portion of this work. I'm not entirely clear how to easily see the results of this work without doing a fair bit of manual tweaking of data.

I'll try to take a look at that later this afternoon but if I can't create a meaningful test set for myself I may defer the "does this work as promised" portion of the check to stakeholders after we merge it as they'll (hopefully) better understand how to get records into a useful state in staging...I hope.

JPrevost (approved):
Updating to approved. I feel like with the tests and code review I understand this as well as I am going to. Let's just extra communicate with stakeholders on the need to get the staging site to have useful data for them to verify these changes before we move it to prod.

PR #1113, opened 2023-02-08
Refactor tests for consented_to_proquest scopes

jazairi:
Why these changes are being introduced:

Tests for the `consented_to_proquest` and `not_consented_to_proquest` scopes alter data in fixtures used for previous assertions rather than using fixtures that specifically meet each test case. This can be brittle, if one were to modify the test without noticing the changes in between assertions.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-621

How this addresses that need:

This adds more fixtures to meet each test case of the ProQuest consent scopes.

Updated a couple of tests to pass, including refactoring an advisor test not to attempt to select a specific thesis attached to a specific user. That caused the test to fail in this case, and the new approach seems more resilient.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/test/fixtures/authors.yml
+++ b/test/fixtures/authors.yml
@@ -121,3 +121,27 @@ eighteen:
   thesis: ready_for_full_export
   graduation_confirmed: true
   proquest_allowed: true
+
+nineteen:
+  user: yo
+  thesis: pq_conflict_true_nil
+  graduation_confirmed: true
+  proquest_allowed: true
+
+twenty:
+  user: basic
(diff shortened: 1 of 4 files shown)


JPrevost (approved):
This is definitely better, but I do think we need to consider supplementing our fixtures with either factories or helpers rather than having a fixture for every condition which is becoming untenable as the app and edge cases expand.

As this is the direction the app has gone so far, :shipit: . One day, we'll have a good engineering discussion where we consider alternatives with this app in mind :)

PR #1119, opened 2023-02-21
Rails 7 upgrade

jazairi:
#### Developer

#### Code Reviewer

NO

YES


db/schema.rb
```diff
@@ -53,16 +52,16 @@
 
   create_table "advisors", force: :cascade do |t|
     t.string "name", null: false
-    t.datetime "created_at", precision: 6, null: false
-    t.datetime "updated_at", precision: 6, null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
```
JPrevost, line 56: I assume this is just from a db:migrate after updating, but do you have any thoughts on why some tables have created_at/updated_at with `precision: nil, null: false` and others have had a previous precision removed and are now  `null: false`?

It seems like things without a precision set previously got the new `precision: nil` added whereas previous `precision: 6` got removed. Maybe it's a new default somewhere to set it to 6? It's making me slightly nervous to not understand why it's not consistent across tables (i.e. I don't particularly care how it works as long as it works but the fact it isn't the same makes me worried something odd is going on...)

jazairi: Removal of `precision: 6` is in fact a new default. Setting precision on timestamps how happens internally in Rails, so it doesn't need to be specified in the schema: https://github.com/rails/rails/blob/v7.0.2.3/activerecord/CHANGELOG.md#rails-702-february-08-2022

Setting `precision` to `nil` is concerning to me. I suspect that was added to the `created_at` columns in the `active_storage_attachments` and `active_storage_blobs` tables because those columns did not explicitly assign precision in the past. I'm not sure that it's intentional, though. I'll do a little more research on that one.

Ideally, this change would have been included in the new framework defaults commit. Either I ran `rails db:migrate` in this commit and forgot that I did, or it happened as part of `rails app:update` somehow. Regardless, I'll document it as a side effect in the final set of commit messages.

JPrevost (approved):
Unless we find that precision nil is an actual concern, this looks good to me.

Let's have stakeholders confirm app functionality in staging before promoting to prod as they will know more than us if we borked something along the way :)

Before handing off to stakeholders we should test that existing holds still display the data we pull directly from paper_trail in the UI. I didn't have any existing holds locally to test this with but I suspect staging will. (I only have a very vague recollection that we do something like this so if it works differently than I remember that's fine too :) )

PR #1222, opened 2023-09-15
Ensure theses are not publishable without an accession number

jazairi:

Because we want the accession number to be used in the preservation workflow, we do not want a thesis without an accession number to be published.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETD-596

This adds an `accession_number` convenience method to the Thesis model that can look up an accession based on a thesis' degree period. This method is included in the publication checks.

The thesis processing form now shows whether it can find an accession number. If no accession number exists, the form will render a hint that prompts the user to create an accession number for the given degree period.

* A future state of this application is for the Thesis model to belong to the Degree Period model. For now, there is no direct assocation, so we are using ActiveRecord lookups based on the thesis' graduation year and graduation month.
* Some fixtures have been added to allow tests related to publication status to pass. (Certain thesis fixtures had no degree periods or accession numbers, they could not be put into publication review.)
* The `new` action on the Archivematica Accessions dashboard has been overwritten to prefill the degree period ID from params.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/thesis.rb
+++ b/app/models/thesis.rb
@@ -146,6 +146,19 @@ class Thesis < ApplicationRecord
 
   enum proquest_exported: ['Not exported', 'Full harvest', 'Partial harvest']
 
+  # Looks up the thesis' accession number based on its degree period.
+  def accession_number
+    degree_period = look_up_degree_period
+    return if degree_period.nil?
+    return if degree_period.archivematica_accession.nil?
+
+    degree_period.archivematica_accession.accession_number
+  end
+
+  def look_up_degree_period
+    DegreePeriod.find_by(grad_year: graduation_year, grad_month: graduation_month)
+  end
+
   # Returns a true/false value (rendered as "yes" or "no") if there are any
   # holds with a status of either 'active' or 'expired'. A false/"No" is
   # only returned if all holds are 'released'.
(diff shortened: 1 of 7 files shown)


app/models/thesis.rb
```diff
@@ -146,6 +146,15 @@ class Thesis < ApplicationRecord
 
   enum proquest_exported: ['Not exported', 'Full harvest', 'Partial harvest']
 
+  # Looks up the thesis' accession number based on its degree period.
+  def accession_number
+    degree_period = DegreePeriod.find_by(grad_year: graduation_year, grad_month: graduation_month)
```
matt-bernhardt, line 151: I intended for this to be a much more thorough review, but that's not looking likely at this point this afternoon. I did want to at least ask this question, with the caveat that I may have forgotten too much of Rails for it to be meaningful:

It makes me a bit nervous to see basically the entire `look_up_degree_period` helper repeated in the Thesis model - is it possible for this to use the helper, rather than replicate it?

JPrevost: You can't directly call a Helper from a Model. This is where some people abstract this out to a Concern.

JPrevost: Or maybe in this case `look_up_degree_period` could actually be in the model itself. I haven't made it deep enough into the review to see how it is used in the UI.

PR #1229, opened 2023-09-25
Use Archivematica accession numbers in SIP AWS keys

jazairi:

Artefactual has requested that our AWS keys for SIPs should include an Archivematica accession number. This will allow them to automate processing of thesis SIPs.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-593

This builds on previous work to rename the AWS key for a given thesis' SIP using the accession number for that thesis.

This will only affect theses preserved _after_ this change is deployed in production. We do not have a plan to retroactively change AWS keys, and stakeholders understand and have agreed to this.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/submission_information_package_zipper.rb
+++ b/app/models/submission_information_package_zipper.rb
@@ -19,7 +19,7 @@ class SubmissionInformationPackageZipper
   # This key needs to be unique. By default, ActiveStorage generates a UUID, but since we want a file path for our
   # Archivematica needs, we are generating a key. We handle uniqueness on the `bag_name` side.
   def keygen(sip)
-    "etdsip/#{sip.thesis.graduation_year}/#{sip.thesis.graduation_month}/#{sip.bag_name}.zip"
+    "etdsip/#{sip.thesis.accession_number}/#{sip.bag_name}.zip"
   end
 
   # bagamatic takes a sip, creates a temporary zip file, and returns that file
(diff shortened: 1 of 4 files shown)


app/models/submission_information_package_zipper.rb
```diff
@@ -19,7 +19,7 @@ def initialize(sip)
   # This key needs to be unique. By default, ActiveStorage generates a UUID, but since we want a file path for our
   # Archivematica needs, we are generating a key. We handle uniqueness on the `bag_name` side.
   def keygen(sip)
-    "etdsip/#{sip.thesis.graduation_year}/#{sip.thesis.graduation_month}/#{sip.bag_name}.zip"
+    "etdsip/#{sip.thesis.accession_number}/#{sip.bag_name}.zip"
```
JPrevost, line 22: The way `thesis.accession_number` works, it is possible for it to return `nil` in situations in which the an accession number was not created or the degree period somehow doesn't exist yet in the application.

I think our expectation is we won't allow publishing to happen and thus sip creation is impossible without valid accession numbers being available. Is that correct and do you think we should do something else here to protect against nil accession numbers?

test/models/submission_information_package_test.rb
```diff
@@ -146,6 +146,19 @@ class SubmissionInformationPackageTest < ActiveSupport::TestCase
     assert_not sip.valid?
   end
 
+  test 'a SIP is invalid if its thesis has accession number' do
```
JPrevost, line 149: The test looks good, the name is backwards. I think it means to be `a SIP is invalid if its thesis has NO accession number`

JPrevost (changes requested):
Minor change requested in a test name, but otherwise this should be good to go.

JPrevost (approved):

PR #1237, opened 2023-10-05
Allow processors to reset publication error status

JPrevost:

* There is no way in the application to reset a publication error status other than manually resaving each thesis that has errored. Until his point, we only had one or two errors per patch so this was not a huge problem. However, recently our vended DSpace ran out of disk space and rejected a batch of 40 theses. Rather than manually resetting each, we felt it was time to add the ability for processors to reset all errors to move them back to publication review.

Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-638

* Creates a Thesis controller restricted to processors and above that changes all Theses with the error status to the review status
* Displays a link to reset publication errors on the view that shows the list of errors

#### Developer

#### Code Reviewer

NO

NO


matt-bernhardt (approved):
This all looks okay to me. I do want to ask one question, though, to make sure that I'm remembering this application correctly. Assuming my memory is correct, then I'm at :shipit: for this.

I see this `reset_all_publication_errors` method is explicitly setting all statuses to "Not ready for publication" - but when I poke on things locally, it looks like that's not actually how theses end up being assigned (which is a good thing - theses which _should_ be in "Publication review" should end up being set to that status as a result of that button). My recollection is that this method is really a way to trigger the `update_status` method, which calculates the appropriate status. `update_status` cannot touch theses which have experienced errors, and it isn't a good idea to directly assign anything to "publication review" status, so we're assigning everything to "not ready" and letting the existing tooling do it's thing.

Assuming all that is right, then I'm fine with this. If I've mis-remembered things, though, I don't want to approve this until I've got it in my head correctly.

JPrevost:
@matt-bernhardt that's a good point and distinction worth noting. We are actually setting the status to what we asked it to be set to in this code, but we have logic that calculates the status and thus records that are _not_ publishable will not be put in the status we asked for. However, anything in Publication Error in a non-local environment should actually end up in the status we ask for because before it could have an error, it had to have been publishable :)

The exception to that would be if we change the thesis publishing requirements _after_ something was moved into the error state. Those would not be put into review and instead would end up in not ready status.

Our initial thought on how to address this was to just resave all of the files which would trigger the same thing, but this felt cleaner to me to explicitly set this to a "not ready" status and let things move to wherever they belong... but to be clear, just resaving ends up with the same result.

PR #1243, opened 2023-10-20
Disable transfers when in maintenance mode

jazairi:

In anticipation of migrating to new AWS infrastructure, we will need to pause file uploads for a bit.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/ETD-649

This adds a maintenance mode feature, toggled via the flipflop gem, that disables the thesis transfer form and changes the welcome message on `transfer/new` to explain what is happening. A custom message can be configured to provide additional details.

* It is still technically possible to post a new transfer if one were to bypass the form. I find it unlikely that transfer submitters would try to hack this, but if it's a concern, we could try a different approach.
* Currently, maintenance mode only disables thesis transfers. I chose to give the feature a more generalized name in case we want to use it to disable additional functionality in the future.

#### Developer

#### Code Reviewer

NO

YES


app/views/transfer/_welcome_maintenance.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="well maintenance-message">
+  <% if ENV.fetch('MAINTENANCE_MESSAGE_TRANSFER', false) %>
+    <h4 class="hd-5"><%= ENV['MAINTENANCE_MESSAGE_TRANSFER'].to_s %></h4>
```
JPrevost, line 3: We should be able to trust this input and making this `.html_safe` instead of `.to_s` would allow more flexibility in the custom messages.

app/views/transfer/new.html.erb
```diff
@@ -173,3 +177,9 @@
       });
     </script>
   <% end %>
+
+  <% if Flipflop.enabled?(:maintenance_mode) %>
+    <script type="text/javascript">
+      $(':input').prop('disabled', true);
```
JPrevost, line 183: Adding
`$('#new_transfer').hide();`
in addition to disabling feels like it would be slightly less confusing to someone that stumbled across this. Thoughts?

jazairi: Yeah, that's a good suggestion. I don't see any downside to hiding the form as well.

JPrevost (approved):
👍🏻

PR #1250, opened 2023-11-21
Refactor error handling in Preservation Submission Job

jazairi:

The Preservation Submission Job currently handles exceptions during SIP creation by setting the SIP's status to 'error' and saving it. This allows us to see failed SIPs in the administrate dashboard, but it also causes preservation to fail silently.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ETD-647

This refactors the Preservation Submission Job to create SIPs using `create!` instead of `create`, so that exceptions are raised during the object creation. When exceptions are raised, they are logged and added to the `results` array. Since no SIP is created when an exception is raised by `create!`, the logic in the `rescue` block around updating `preservation_status` has been removed.

* Fixtures have been adjusted to accommodate testing.
* A quasi-related test that had no assertions has been updated to actually test something.
* The `preservation_status` of 'error' has been removed, as it is no longer used in the app. There are currently no SIPs in production that have this status.
* Unrelated to this change, the `baggable` module has been moved from `app/models` to `app/lib`. This to be consistent with where we store our other module, and to make it clear that it is not a model.

#### Developer

#### Code Reviewer

NO

NO


app/jobs/preservation_submission_job.rb
```diff
@@ -6,15 +6,13 @@ def perform(theses)
     results = { total: theses.count, processed: 0, errors: [] }
     theses.each do |thesis|
       Rails.logger.info("Thesis #{thesis.id} is now being prepared for preservation")
-      sip = thesis.submission_information_packages.create
+      sip = thesis.submission_information_packages.create!
       preserve_sip(sip)
       Rails.logger.info("Thesis #{thesis.id} has been sent to preservation")
       results[:processed] += 1
     rescue StandardError, Aws::Errors => e
```
JPrevost, line 13: Hmmm. Does `create!` failing cause the rescue block to be triggered? If so, the `error` status would still work right? If not, should we update the rescue to also catch the specific  exception being throw so we get a useful `results[:errors]` list?

jazairi: `create!` failing does trigger the `rescue` block. However, the `error` status does not work in this case because an object is not created when `create!` fails, thus there is no SIP of which we can change the `preservation_status`.

JPrevost: lol...how many times will you have to explain this to me? I feel like it's at least 4 so far. <sorry!>

JPrevost (approved):

PR #1310, opened 2024-05-01
Add additional context to a note about `pending publication`

JPrevost:
There is no code, only a docs update to provide additional context to an edge case we are now supporting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -390,7 +390,7 @@ task:
 heroku run -s standard-2x rails dss:publish_thesis_by_id[THESIS_ID] --app TARGET-HEROKU-APP
 ```
 
-Note: `Pending publication` is allowed here, but not expected to be a normal occurence, to handle the edge case of the app thinking data was sent to SQS but the data not arriving for any reason.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -390,7 +390,7 @@ task:
 heroku run -s standard-2x rails dss:publish_thesis_by_id[THESIS_ID] --app TARGET-HEROKU-APP
 ```
 
-Note: `Pending publication` is allowed here, but not expected to be a normal occurence, to handle the edge case of the app thinking data was sent to SQS but the data not arriving for any reason.
+Warning: `Pending publication` is allowed here, but not expected to be a normal occurence, to handle the edge case of the app thinking data was sent to SQS but the data not arriving for any reason. It is possible to send multiple copies of a single thesis to the publishing queue with this feature. Please be sure the SQS queues are fully processed and that you are absolutely confident that sending a `Pending publication` status thesis back into the publishing queue is appropriate. Failure to use care could lead to a single thesis being published more than once, which is bad. This is only a concern specifically for the `Pending publication` and `Publication review` statuses as no other statuses can be sent to the queue with this method.
```
jazairi, line 393: What are the implications of double-publishing? Multiple DSpace records?

jazairi (changes requested):
Could you include information about the consequences of publishing the same thesis multiple times? I assume the issue is duplicate DSpace records and/or multiple handles, but it may be useful to explicitly document what the risk is.

README.md
JPrevost: Ha. Great question and I agree it should be added here for context. I'm not actually even sure what would happen in our app in this situation, but it would definitely involve having the thesis team clean up the extra record. I'll leave this open until I can figure out how best to word the important context here.

jazairi: Fwiw, I think it would provide sufficient context to say something like, "which may result in duplicated theses being published and other unexpected consequences." (I don't think we necessarily need to document _everything_ that could go wrong.)

jazairi:
@JPrevost I noticed this PR while doing some maintenance and tried to do a bit more wordsmithing. Let me know what you think!

JPrevost (commented):
That seems very clear. Thanks for taking the time to explain the situation better!

PR #1328, opened 2024-07-08
Rollback AWS gems

JPrevost:
- the gems have changed an internal protocol that is breaking some tests. We need to rollback to a known working version to resolve production issues and then get to a point where we can upgrade all AWS gems to the latest version.

https://mitlibraries.atlassian.net/browse/ENGX-276

NO

YES

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,8 +7,8 @@ git_source(:github) do |repo_name|
 end
 
 gem 'administrate', '~> 0.19.0'
-gem 'aws-sdk-rails'
-gem 'aws-sdk-s3'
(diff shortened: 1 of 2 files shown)


jazairi (approved):
Confirmed that this matches the last working versions, from before the most recent rounds of dependency updates. I'm not convinced that S3 breaks in more recent versions, but I think it's reasonable to pin it to avoid any additional issues. Similarly, while it's likely that aws-sdk-sqs will work on 1.63.0, I have no problem pinning to 1.62.0 for the time being.

PR #1329, opened 2024-07-09
Updates all AWS gems and skips some assertions and tests

JPrevost:
Our tests have some syntax errors in some of the mocks. We will fix them, but for now we are skipping failures and pushing the updated gems to allow us to read the production SQS queues which is not currently possible due to a change introduced somewhere in SQS or the associated gems.

NO

YES


test/jobs/dspace_publication_results_job_test.rb
```diff
@@ -26,7 +26,7 @@ def setup
                                       'SubmissionSource' => { string_value: 'ETD', data_type: 'String' } } },
 
               # success but thesis no longer has files locally
-              { message_id: 'id1a', receipt_handle: 'handle1a',
+              { message_id: 'id1b', receipt_handle: 'handle1a',
```
JPrevost, line 29: I... lol. Great catch!

jazairi (approved):
See the commit I pushed for a possible solution to the failing test issue. I'm approving this with or without that commit, as I think it's safe to skip those tests based on my current understanding.

Thank you for taking the time to solve this and upgrade these long-outdated gems!

test/jobs/dspace_publication_results_job_test.rb
jazairi: Yeah, I felt pretty silly when I caught that. 😅 (Both that we had duplicate IDs, and that we chose such a weird, brittle naming pattern!)

PR #1396, opened 2024-12-03
Remove ProQuest submission field from thesis forms

jazairi:

The Libraries are suspending the pilot with ProQuest, and the thesis forms should be updated.

#### Relevant ticket(s):

* [ETD-658](https://mitlibraries.atlassian.net/browse/ETD-658)

This removes the ProQuest fieldset from the new and edit forms. It also removes the author_fields partial, which includes only this fieldset.

Note that this does not remove any additional ProQuest export functionality, as specifically requested by stakeholders.

Because no functionality outside of the thesis form has been changed, thesis processors can still run ProQuest exports. There is no apparent risk to this, and we can fully remove the feature if stakeholders decide to do so.

#### Developer

#### Code Reviewer

NO

NO

[ETD-658]: https://mitlibraries.atlassian.net/browse/ETD-658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JPrevost (approved):
This looks good. I forgot how complex this feature is under the hood and it took me a while to confirm the Export would work as I expected with the `null` values we'll have (the tests had coverage for it which was great, but I wanted to _see_ it working locally (and it worked differently than I expected but is doing the correct thing so I'm glad I checked). After further reflection, it makes sense it works the way it does as Users aren't actually required to login and update their metadata so `null` has always been a supported status for this.

Please tag Mikki in the ETD channel and Jira ticket to review once this merges but before it is moved to production. Production launch for this will be whenever the ETD says to launch it (not sure if they want to update some supporting instructions first or if they'll want this live once they confirm it).

JPrevost:
Also, I noticed we don't update the strong params (https://github.com/MITLibraries/thing/blob/main/app/controllers/thesis_controller.rb#L193) as part of this but I think because the feature is remaining on the admin-side it makes sense to leave it as-is. I'm saying that here in case you look at it and disagree and think it can come out now. We seem to use the same params for multiple form updates which seems maybe not the best idea but is outside the scope of this changeset.

PR #1399, opened 2024-12-06
Make requested changes to the copyright and license fields

jazairi:

Stakeholders have requested that we preselect the 'author' copyright value, and make the license field conditionally required when that copyright value is selected.

#### Relevant ticket(s):

* [ETD-627](https://mitlibraries.atlassian.net/browse/ETD-627)

This adds the `selected` property to the desired option in the copyright field. It also updates the `conditionalLicenseField` function to toggle the `required` property on the license field when it is shown or hidden.

* Past experience has made me wary of toggling the `required` prop in this way. However, it is generally reliable under a single binary condition (either the 'Author' copyright value is selected, or it isn't). I haven't noticed any issues in click-testing.
* `selected` is tied to a value, so in this case we are preselecting the first option. This corresponds to 'Author' in both staging and prod, but it would be better to make an explicit association. If the data were to change order somehow, the wrong copyright holder could be preselected.

#### Developer

#### Code Reviewer

NO

NO

[ETD-627]: https://mitlibraries.atlassian.net/browse/ETD-627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/thesis/new.html.erb
+++ b/app/views/thesis/new.html.erb
@@ -124,6 +124,7 @@
     </div>
 
     <%= f.association :copyright, as: :select,
+                                  selected: 2,
                                   required: true,
                                   collection: Copyright.display_to_author,
                                   validate: { presence: true },
@@ -140,12 +141,13 @@
 
     <%= f.association :license, as: :select,
                                 include_hidden: false,
-                                label: 'License (if you retain copyright)',
+                                label: 'License (if you retain copyright) *',
                                 label_method: :display_description,
                                 label_html: { class: 'col1q' },
                                 wrapper_html: { class: 'field-row select layout-1q3q layout-band' },
                                 input_html: { class: 'field field-select col3q', multiple: false,
-                                  aria: { describedby: 'thesis_license-hint' } },
+                                  aria: { describedby: 'thesis_license-hint' },
+                                  data: { msg: Thesis::VALIDATION_MSGS[:license] } },
                                 hint: 'Not sure what to select? Learn more about <a href="https://chooser-beta.creativecommons.org/" target="_blank">Creative Commons licenses</a>.'.html_safe,
                                 hint_html: { class: 'col3q', id: 'thesis_license-hint' } %>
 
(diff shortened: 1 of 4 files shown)


app/views/thesis/new.html.erb
```diff
@@ -124,6 +124,7 @@
     </div>
 
     <%= f.association :copyright, as: :select,
+                                  selected: 1,
```
JPrevost, line 127: @jazairi It looks like we set the ID is consistent in the PR builds as well so while not ideal, it's probably a risk we can accept as there is little reason we'd be changing the Author copyright to be something inherently different under the good.

That said, I believe this ID is wrong as Author is ID 2 (in the PR build and prod at least).

jazairi: @JPrevost Thanks for pointing that out! While fixing that, I noticed that I'd forgotten to include license validation/required field indicator in the edit view, so I added that, too.

app/views/thesis/new.html.erb
```diff
@@ -124,6 +124,7 @@
     </div>
 
     <%= f.association :copyright, as: :select,
+                                  selected: 2,
```
JPrevost, line 127: I wonder if for the edit view we can do something like `selected: 2 if license.nil?` (not real code). My reasoning is that because of the registrar data load pre-creating theses, I think most users hit the edit form and not the new form.

jazairi: Ah, good point. I'll see what options we have there.

jazairi:
@JPrevost I think the latest commit addresses your concern about the edit form. It's a bit kludgy to preselect the associated copyright, but without that condition, that field would be blank if the thesis already had a copyright.

JPrevost (approved):
This works how I think it makes sense... let's hope when you check in with stakeholders it's what they meant :) 

Thanks for taking the extra time to make the tweaks!

PR #1426, opened 2025-02-24
Rails71

JPrevost:

YES

YES


jazairi (approved):
Thanks for picking this up, and thanks for the updates about this work in Slack and the ticket! It was helpful to come into this review with some awareness of your process.

While I'd prefer to be using 7.1 framework defaults, I agree that running 7.1 with 7.0 defaults is a significant improvement for maintenance purposes. I think the key will be remembering that we're not using 7.1 defaults, but I think that leaving the `new_framework_defaults` file in the repo should help with that.

That db migration made me a tad nervous, too, so I like your idea of backing up prod prior to promotion.

Otherwise, everything here is consistent with my expectations. (It will always be hilarious to me that rubocop and Rails disagree on single quotes vs double quotes.)

:shipit:

PR #1428, opened 2025-02-27
Update deps

JPrevost:
Please confirm the Hold logic (New Hold from a Thesis and Thesis summary panel in a Hold) look and function as expected as the main focus of this review. I believe I was able to integrate upstream changes and overlay our local changes to result in the expected features... but I'd really appreciate someone checking this type of work to confirm I didn't miss something.

#### Developer

#### Code Reviewer

NO

YES


app/views/admin/theses/show.html.erb
```diff
@@ -30,6 +30,13 @@ as well as a link to its edit page.
       [:edit, namespace, page.resource],
       class: "button",
     ) if accessible_action?(page.resource, :edit) %>
+    <%= link_to(
```
jazairi, line 33: I have some concern about adding this button back in. I suspect there's a reason it's not there in the first place. That said, I trust Wilder not to delete anything that shouldn't be deleted, so if you're comfortable with it, I think it's fine. (At the very least, it will make future upgrades less confusing.)

jazairi (commented):
You seem to have fixed the weird container overflow bug that was annoying me last time. Kudos!

I pushed a small change to the markup to put the new hold button inline with the others. Let me know what you think of it. We can add more spacing if necessary, but the misalignment was bothering me.

app/views/admin/theses/show.html.erb
JPrevost: I couldn't actually find where it was in our code ever and assumed it was new to the template honestly.

As it's conditional as to whether it displays, maybe we check in with Wilder as to whether they can see this button... and if they can see it, adjust permissions until they can't?

JPrevost:
@jazairi thanks for fixing the alignment. I hated it but was too lazy to track down why it was weird. 🍻

app/views/admin/theses/show.html.erb
jazairi: Yeah, I think you're right. I see the destroy button in that view for the [v0.19 release commit](https://github.com/thoughtbot/administrate/blob/482bc010581440d289074ff454c85cff4e65c386/app/views/administrate/application/show.html.erb#L33-L40), but _not_ in some earlier releases. It seems they added it at some point, and we just didn't pull it in to the customized template. Knowing that it likely wasn't an intentional customization makes me feel better about keeping it.

Incidentally, I confirmed that it's visible to the `thesis_admin` role, so Wilder should be able to see it.

jazairi (approved):

PR #1476, opened 2025-10-08
Rails 7.1 framework defaults

JPrevost:
One way to see this change:

* Locally, create a new transfer with a few files
* In a rails console
  * `t = Transfer.last`
  * `t.files_count`
  * `t.unassigned_files_count`

If everything is working as expected, both file counts should match the number of files you attached in the UI. 

To see that this was fixed, revert line 98 in `app/models/transfer.rb` to `save`, restart your server and console and run the same test. You should get 0 `unassigned_files_count` which is incorrect.

https://mitlibraries.atlassian.net/browse/ETD-66

NO

NO

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,5 @@
 AllCops:
-  TargetRubyVersion: 2.7
+  TargetRubyVersion: 3.2
   NewCops: enable
   StyleGuideBaseURL: https://rubystyle.guide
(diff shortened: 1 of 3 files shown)


jazairi (approved):
Confirmed locally that this fixes the problem by running the test described in the PR description. Thanks for figuring out a long-standing problem!

